PR #47, opened 2013-09-20
Support commonjs module scheme.

btakita:
Fix for https://pusher.tenderapp.com/discussions/requests/1004-support-nodejs-jsdom-client-side-testing.

I updated the logic to protect against the module global variable being set. exports, module, & module.exports need to be truthy for this to apply.

### DIFF
--- a/src/pusher.js
+++ b/src/pusher.js
@@ -81,6 +81,12 @@
 
     if (Pusher.isReady) self.connect();
   }
+  if ( typeof module === "object" && module && typeof module.exports === "object" ) {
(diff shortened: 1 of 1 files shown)


marcbachmann:
Any updates? Can we help somehow?

leggetter:
Part of resolving #84 will also resolve this as a UMD wrapper will be added as part of the build process.

I'll update the status of that issue so please subscribe to that issue for updates.

leggetter:
Closing since 3.0 is now on our CDN and accessible via Bower and NPM.

PR #91, opened 2014-12-10
Customisable xhr

lachie:
Add an optional callback to `Pusher`: `Pusher.buildXHR`.

`Pusher.XHR` exists, but needs to be a constructor (its used like `xhr = new Pusher.XHR();`)

I just wanted to write a function which passed back an XHR instance:
```javascript
xhr = Pusher.buildXHR();
```

I opted to add this function so as not to break the API.

Note that I also fixed up the build script so that it worked on my greenfields machine (ubuntu, not mac)


lachie:
Thanks, sorry about that. I commented in the other PR

PR #101, opened 2015-02-22
Unixify build scripts

lachie:
I needed to write a patch for pusher.js, but found I couldn't get it to build on my linux dev machine. Here are the changes I made to get going!

- fixes filename case for case-sensitive FS
- makes build script more robust
- improved build script for web-socket-js

(Note: I closed and opened a new PR after rebasing :)


pl:
Thanks a lot, working on these scripts is always a pain. We're removing Flash in the next major release, so it will simplify the build a lot.

I'll merge it to master, though I think I will skip the Jfile renaming and fix jbundle instead. It seems that the intention of the author was to call the file `JFile` and since nobody used the tool with a case-sensitive FS, it went unnoticed.

zimbatm:
Thanks @lachie , I've integrated some of your changes in b685d4d72e78d55fdbc72ed4c155fcff62eb878a for the next major release

PR #102, opened 2015-02-22
Build a custom xhr object

lachie:
I needed to supply a new XHR object, rather than a prototype to be instantiated.

The use case is something like this:
```javascript
Pusher.buildXHR = function() {
  xhr = new XMLHttpRequest();
  xhr.withCredentials = true;
  return xhr;
}
```

Note that this is built on top of PR #101 . Happy to rebase this if you end up merging #101 first.


pl:
Hi Lachie,

You can use `Pusher.XHR` exactly for this - the return value of the function is the result of the `new` statement if it's an object:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new

lachie:
Hi @pl sorry I missed your response to my last PR (forgot I even raised it!).

I tried doing it the way you suggested initially, but couldn't get it working... the error was about constructors. Now I'm trying it and its fine... :anguished:

Anyway, at least I cleaned up the build scripts in the process :)

PR #109, opened 2015-05-30
Allow to forgo the `new` keyword

zimbatm:
Don't break in mysterious way when the user forgets to add the `new`
keyword like this:

    var pusher = Pusher("key");

### DIFF
--- a/src/pusher.js
+++ b/src/pusher.js
@@ -1,5 +1,9 @@
 ;(function() {
   function Pusher(app_key, options) {
+    if (!this instanceof Pusher) {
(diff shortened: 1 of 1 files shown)


leggetter:
Pros/cons over throwing an exception?

zimbatm:
Actually I would prefer the exception, I just didn't had a good description of the error to put in the exception message.

leggetter:
>Actually I would prefer the exception, I just didn't had a good description of the error to put in the exception message.

I'd vote to back this out, do a `instanceof this` check and come up with a clear exception message.

PR #111, opened 2015-06-12
Introduce simplified debug logging

zimbatm:
This changes intends to simplify the recommended way for pusher-js
to enable logging.

The previously recommended way:

    Pusher.log = function(message) {
      if (window.console && window.console.log) {
        window.console.log(message);
      }
    };

The new recommended way:

    Pusher.verbose = true;

This change is backward-compatible. If someone installs his own logger
then `Pusher.verbose` is simply ignored.

In old IE `Pusher.verbose` is also ignored.

### DIFF
--- a/src/pusher.js
+++ b/src/pusher.js
@@ -90,6 +90,16 @@
   Pusher.instances = [];
   Pusher.isReady = false;
 
+  Pusher.verbose = false;
+
+  if (window.console && window.console.log) {
+    Pusher.log = function(message) {
+      if (Pusher.verbose) {
+        window.console.log(message);
+      }
+    };
+  }
+
   // To receive log output provide a Pusher.log function, for example
   // Pusher.log = function(m){console.log(m)}
   Pusher.debug = function() {


hamchapman:
:+1: I'm in favour of this.

@leggetter WDYT?

leggetter:
I'm definitely in favour of us making it much easier to turn logging on and off.

I believe we've been moving away from global configuration. But I understand that logging my be something we're going to have to keep as global.

`verbose` is a logging level and potentially suggests we have other levels that you can configure logging for, which isn't the case.

It's really either logging on or logging off. Maybe `logToConsole`. Or `logging` where `true` and `false` mean on/off.

zimbatm:
Yeah it would be nice to have it per Pusher instance. The diff would be a bit bigger to pass the log function around.

We currently have two log-levels: debug and warning. Pusher.log is used by Pusher.debug().
Pusher.warning tries to use console.warn and console.log even if Pusher.log is not defined: https://github.com/pusher/pusher-js/commit/64a4daa3162a66d990ed1f15cfa263a3cfeb767a#diff-38d6124aece475ccbc67619712b4149aL102

There is also a Pushr

leggetter:
IMO the use of `verbose` is confusing. My vote is to back this out and choose a better name or logging strategy with some references around why that name or strategy has been chosen.

PR #117, opened 2015-08-29
Rename ambiguous Pusher.verbose to Pusher.logToConsole. Add some tests.

leggetter:


leggetter:
@zimbatm Renamed to `Pusher.logToConsole` https://github.com/pusher/pusher-js/commit/52e10340411e03ddcabe7d94e5875c17fe1373cf

Usage will now be:

```html
<script src="//js.pusher.com/3.0/pusher.min.js"></script>
<script>
  Pusher.logToConsole = true;

  var pusher = new Pusher('APP_KEY');
</script>
```

README.markdown
```diff
@@ -24,31 +25,48 @@ The following topics are covered:
 
 ## Installation
 
-Via the Pusher CDN (or [CDNJS](https://cdnjs.com/libraries/pusher)):
```
zimbatm: Was CDNJS removed on purpose ?

zimbatm:
@leggetter thanks for pushing this through, `logToConsole` is a good compromise.

Last thing would be to restore the CDNJS reference and remove the unnecessary intermediate commits. I am happy to do it, just waiting for your approval so we don't both force-push on the branch.

leggetter:
I don't think we should add the CDNJS reference back. 

As part of [the merge](https://github.com/pusher/pusher-js/commit/37b755fc68c38626cd7797549496bcf87b89eed8) I left this comment:

> If we are going to list CDNJS then we should push updates to it as part of the release process. Since that’s not part of the release process right now it’s being removed.

README.markdown
leggetter: Yeah, please seem comment below.

leggetter:
Replaced by #121

PR #121, opened 2015-09-19
Introduce simplified debug logging

leggetter:
Replaces #117. 

This changes intends to simplify the recommended way for pusher-js
to enable logging.

The previously recommended way:

    Pusher.log = function(message) {
      if (window.console && window.console.log) {
        window.console.log(message);
      }
    };

The new recommended way:

    Pusher.logToConsole = true;

This change is backward-compatible. If someone installs his own logger
then `Pusher.logToConsole` is simply ignored.

### DIFF
--- a/CHANGELOG.markdown
+++ b/CHANGELOG.markdown
@@ -1,5 +1,9 @@
 # Changelog
 
+## 3.1.0-pre
+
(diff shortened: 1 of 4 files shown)


leggetter:
@zimbatm - I've attempted a rebase :blush:.

PR #128, opened 2015-10-04
[[Speculative]] Abstract `window` and `document` behind a `Pusher.runtime` object

jpatel531:
Still untidy and a work in progress and is meant to be a piece for discussion rather than anything else. It came about when I tried to get PusherJS working for ReactNative, but the ReactNative runtime does not have a `document`. @leggetter recommended I try to get PusherJS working in a WebWorker to test things out. WebWorkers, similarly, do not have a `document`, nor does it have a `window`.

This basically detects the runtime environment, i.e. `Browser` or `NonBrowser`. Each `runtime` has `getWindow` and `getDocument` method. I've replaced all direct references to `window` and `document` to `Pusher.runtime.getWindow()` and `Pusher.runtime.getDocument()` respectively. As a result, the library can supply an appropriate value for `window` and `document` based on the environment.

At the moment this is _very_ hacky. There's no appropriate "mock" for `getDocument ` in a non-browser environment - still trying to think about that. PusherJS works in a WebWorker with these changes, and the following code:

```js
importScripts("pusher.js");

var pusher = new Pusher('app_key', {
  disableStats: true
});

var channel = pusher.subscribe('test_channel');

channel.bind('my_event', function(data) {
  postMessage(data); // send the data to the application thread
});
```

WDYT @pl @leggetter @zimbatm ? Is it at least the right start?

### DIFF
--- a/Jfile
+++ b/Jfile
@@ -18,7 +18,7 @@ bundle 'pusher.js' do
   license 'pusher-licence.js'
 
   file 'build/umd_header.js'
-
+  file 'runtime.js'
   file 'pusher.js'
 
   file 'utils/timers.js'
--- a/dist/sockjs.js
+++ b/dist/sockjs.js
@@ -51,7 +51,7 @@ var JSON;JSON||(JSON={}),function(){function str(a,b){var c,d,e,f,g=gap,h,i=b[a]
 //     [*] Including lib/index.js
 // Public object
 var SockJS = (function(){
-              var _document = document;
+              var _document = Pusher.runtime.getDocument();
               var _window = window;
               var utils = {};
 
--- a/src/dependencies.js
+++ b/src/dependencies.js
@@ -16,7 +16,7 @@
 
   // Allows calling a function when the document body is available
    function onDocumentBody(callback) {
-    if (document.body) {
+    if (Pusher.runtime.getDocument().body) {
       callback();
     } else {
       setTimeout(function() {
@@ -29,9 +29,13 @@
     onDocumentBody(initialize);
   }
 
-  if (!window.JSON) {
+  if (typeof(Pusher.runtime.getWindow().JSON)=== undefined) {
     Pusher.Dependencies.load("json2", {}, initializeOnDocumentBody);
   } else {
-    initializeOnDocumentBody();
+    if (Pusher.runtime instanceof _PusherRuntimes.Browser){
+      initializeOnDocumentBody();
+    } else {
+      initialize();
+    }
   }
 })();
--- a/src/dependency_loader.js
+++ b/src/dependency_loader.js
@@ -62,7 +62,7 @@
    */
   prototype.getRoot = function(options) {
     var cdn;
-    var protocol = Pusher.Util.getDocument().location.protocol;
+    var protocol = Pusher.runtime.getDocument().location.protocol;
     if ((options && options.encrypted) || protocol === "https:") {
       cdn = this.options.cdn_https;
     } else {
--- a/src/dom/script_request.js
+++ b/src/dom/script_request.js
@@ -13,6 +13,8 @@
   }
   var prototype = ScriptRequest.prototype;
 
+  var _document = Pusher.runtime.getDocument();
+
   /** Sends the actual script request.
    *
    * @param {ScriptReceiver} receiver
@@ -21,7 +23,7 @@
     var self = this;
     var errorString = "Error loading " + self.src;
 
-    self.script = document.createElement("script");
+    self.script = _document.createElement("script");
     self.script.id = receiver.id;
     self.script.src = self.src;
     self.script.type = "text/javascript";
@@ -44,9 +46,9 @@
     }
 
     // Opera<11.6 hack for missing onerror callback
-    if (self.script.async === undefined && document.attachEvent &&
+    if (self.script.async === undefined && _document.attachEvent &&
         /opera/i.test(navigator.userAgent)) {
-      self.errorScript = document.createElement("script");
+      self.errorScript = _document.createElement("script");
       self.errorScript.id = receiver.id + "_error";
       self.errorScript.text = receiver.name + "('" + errorString + "');";
       self.script.async = self.errorScript.async = false;
@@ -54,7 +56,7 @@
       self.script.async = true;
     }
 
-    var head = document.getElementsByTagName('head')[0];
+    var head = _document.getElementsByTagName('head')[0];
     head.insertBefore(self.script, head.firstChild);
     if (self.errorScript) {
       head.insertBefore(self.errorScript, self.script.nextSibling);
--- a/src/events_dispatcher.js
+++ b/src/events_dispatcher.js
@@ -1,4 +1,5 @@
 ;(function() {
+  var _window = Pusher.runtime.getWindow();
   /** Manages callback bindings and event emitting.
    *
    * @param Function failThrough called when no listeners are bound to an event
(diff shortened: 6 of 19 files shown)


jpatel531:
The more relevant files are `runtime.js`, `dependencies.js` and `pusher.js`

jpatel531:
Also I guess the only transport allowed would be ws

src/dependencies.js
```diff
@@ -16,7 +16,7 @@
 
   // Allows calling a function when the document body is available
    function onDocumentBody(callback) {
-    if (document.body) {
+    if (Pusher.runtime.getDocument().body) {
```
leggetter: This could be moved into the runtime:

```js
if (Pusher.Runtime.isReady()) {
   ...
```

The browser implementation could do the `.body` check.

src/dependencies.js
```diff
@@ -29,9 +29,13 @@
     onDocumentBody(initialize);
   }
 
-  if (!window.JSON) {
+  if (typeof(Pusher.runtime.getWindow().JSON)=== undefined) {
```
leggetter: We could move the fetching of `JSON` from the runtime instead.

Hmmm. Maybe even the dependencies should be loaded by the runtime.

Also - does this work? Here we're checking the value of a string against `undefined`. AFAIK that will always fail. Shouldn't it just be `Pusher.runtime.getWindow().JSON === undefined` ?

src/dependency_loader.js
```diff
@@ -62,7 +62,7 @@
    */
   prototype.getRoot = function(options) {
     var cdn;
-    var protocol = Pusher.Util.getDocument().location.protocol;
+    var protocol = Pusher.runtime.getDocument().location.protocol;
```
leggetter: Could we instead have a `Pusher.Runtime.getProtocol()` ?

src/dom/script_request.js
```diff
@@ -13,6 +13,8 @@
   }
   var prototype = ScriptRequest.prototype;
 
+  var _document = Pusher.runtime.getDocument();
```
leggetter: Doing a request for a file via dynamically creating a `script` tag feels browser runtime specific. So maybe this would be moved into `Pusher.Runtime` if we proceed?

src/events_dispatcher.js
```diff
@@ -1,4 +1,5 @@
 ;(function() {
+  var _window = Pusher.runtime.getWindow();
```
leggetter: In this case we actually want the runtime global. So maybe we should have a `Pusher.Runtime.getGlobal()` where the browser implementation returns `window`?

src/net_info.js
```diff
@@ -5,16 +5,19 @@
    * - online - when browser goes online
    * - offline - when browser goes offline
    */
+
+  var _window = Pusher.runtime.getWindow();
```
leggetter: Net info feels like it should be runtime specific. Certainly the way this information will be detected will be runtime dependent.

src/dependency_loader.js
jpatel531: yeah good idea

src/runtime.js
```diff
@@ -0,0 +1,38 @@
...
+  }
+
+  NonBrowserRuntime.prototype = new Runtime();
+  NonBrowserRuntime.prototype.constructor = NonBrowserRuntime;
+
+  NonBrowserRuntime.prototype.getDocument = function(){
+    return this.context;
+  }
+
+  this._PusherRuntimes = {
```
leggetter: Nitpick: The underscore is probably inconsistent with the existing coding style.

dist/sockjs.js
```diff
@@ -51,7 +51,7 @@ var JSON;JSON||(JSON={}),function(){function str(a,b){var c,d,e,f,g=gap,h,i=b[a]
 //     [*] Including lib/index.js
 // Public object
 var SockJS = (function(){
-              var _document = document;
+              var _document = Pusher.runtime.getDocument();
```
leggetter: This is a built file so the changes would be lost when a release were built.

Right now Pusher knows about SockJS, but not the other way around. It's probably worth trying to come up with an alternative solution.

src/dependencies.js
jpatel531: Oh yeah, it should be `'undefined`. Will fix

Jfile
```diff
@@ -18,7 +18,7 @@ bundle 'pusher.js' do
   license 'pusher-licence.js'
 
   file 'build/umd_header.js'
-
+  file 'runtime.js'
```
leggetter: `runtime.js` should probably be included after `pusher.js`.

`pusher.js` defines the `Pusher` global. So if the contents of `runtime.js` were executed prior to the code in `pusher.js` then the `Pusher` global will not yet exist.

leggetter:
For consistency with other objects `Pusher.runtime` should be `Pusher.Runtime` (kick off a global find & replace :blush:).

src/runtime.js
leggetter: Ah, these don't live on the `Pusher` global. But they probably should as we don't want to pollute the global namespace. Also see my comment about `runtime.js` coming before/after `pusher.js`. Probably a few things that need to be considered here.

src/runtime.js
```diff
@@ -0,0 +1,38 @@
...
+  function NonBrowserRuntime(context){
+    this.context = context;
+  }
+
+  Runtime.prototype.getWindow = function() {
+    return this.context;
+  };
+
+  BrowserRuntime.prototype = new Runtime();
+  BrowserRuntime.prototype.constructor = BrowserRuntime;
```
leggetter: We have a `Pusher.Util` function for doing inheritance. Maybe this was done here because of the loading order? For consistency we should try to use the existing function.

zimbatm:
Instead of introducing a runtime object. In most context you could setup the window object like this: `var global = this;` or `var global = (function(){ return this; })();`. Then for the rest, use or create individual feature-detection functions, I believe there are already a couple of those.

pl:
IMHO, instead of adding abstraction, I'd think whether we actually need to use those features.

1. does the dependency loader work at all in a non-browser context? do we need it?
2. can we submit metrics using a different mechanism?
3. do we need SockJS in non-browser environments?

I'm fairly certain than in Node you don't need SockJS, as you can use a normal HTTP library. My intuition is that React doesn't support browsers that need SockJS (basically those that lack XHR). Those browsers must support JSON too. Our XHR implementation is not that big (around the size of the dependency loader + JSONP code), so we could potentially include it in the main file for those environments.

Submitting metrics via XHR (instead of horrible JSONP) would require exposing a new endpoint, but that's not a big job, IMHO.

jpatel531:
As far as I can tell, the dependency loader wouldn't work non-browser - i.e. an environment without a DOM. Also, certain platforms - such as Atom (not really a target environment but I thought I'd mention it) - with a content security policy directive will reject loading dependencies via the DOM. 

Runtimes such as ReactNative require the Android API to be at least at version 23 (the latest), and newer iOS versions - where WebSocket will be supported. If a browser has WebWorker support it will also have WebSocket support. Am I forgetting other kinds of non-browser environments? 

I think submitting metrics via a different method would be nice - not that I'm too familiar with this area. It may also remove instances where people using AdBlockPlus/uBlock get console errors due to  stats.pusher.com being blacklisted.

pl:
Sounds like we'd be alright in those environments, the question is whether we could merge that changes into master without breaking any browser compatibility. 

We'd still need to support HTTP fallbacks for bad networks, but those environments should provide an HTTP client in form of XHR or something else. Ditching JSONP would be really nice.

PR #130, opened 2015-11-04
Document the unsubscribe method.

jackfranklin:
@zimbatm could you give it a quick check please?


leggetter:
I've ended up here so I've had a quick check and it's good to go.

PR #142, opened 2016-02-19
Throw when app_id is not specified

mdpye:
Warnings are too easily ignored, and this one will definitely become an
error once we try to connect, so there's no point continuing


zimbatm:
Good idea

PR #144, opened 2016-03-09
Better documentation of bind_all methods

jameshfisher:
Following support request for documentation on how to use bind_all

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -276,7 +276,27 @@ Remove previously-bound handlers from an object. Only handlers that match all of
 
 ### Binding to everything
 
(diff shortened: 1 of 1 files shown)


jameshfisher:
(When merging, we should also update the docs pages with this)

jpatel531:
Sounds good - I'll do it tomorrow :)

jackfranklin:
@jpatel531 bump!

jackfranklin:
I should have mentioned this earlier, but long term I'd love to move from `bind_all` to `bindAll` for all snake case methods. We could do this in a way that could add a console warning but not be a breaking change.

PR #150, opened 2016-03-15
update build instructions

jameshfisher:


jackfranklin:
Longer term we should also just rework how we build - should probably move to webpack @jpatel531 ?

jpatel531:
@jackfranklin - yeah i'd love webpack. the problem is that it adds a fair bit to the library size. there are probably a number of stategies though for this but we'd have to look into it.

PR #152, opened 2016-03-24
Pattern matching on event names. Resolves #135

### LINKED ISSUE #135: [REQUEST] Event listener wildcards
Title says it all, I'd like to be able to be able to listen to events like `*:created` or `user:*`. Maybe allow regex?

Or would it just be better to `bind_all` and then manually filter by event name?

jpatel531:
I'm not 100% sure about this but I'm opening this to be open to suggestions.

This PR allows you to pass a regular expression to the `Channel::bind` method. Like so:

```js
channel.bind(/events-/, function(data, match) {
  console.log("MESSAGE!!", data, match);
});
```

On master, the `CallbacksRegistry` stores callbacks in a dictionary `{<key>: [<callback>]}`, where the key is `"_" + eventName`.

Instead of storing a prefixed eventName , this implementation stores a prefixed _pattern_ as the key. When we do lookups on callbacks (when we emit events), we `test` if the emitted event name matches a given key's pattern.  

My main concern is that performing regex tests on keys in the callbacks registry might affect performance. I'm still yet to think of clean ways of approaching this, though. Wanted to get feedback before I start writing tests for it

Happy to take suggestions @zimbatm @jackfranklin

### DIFF
--- a/src/events_dispatcher.js
+++ b/src/events_dispatcher.js
@@ -40,7 +40,7 @@
     var callbacks = this.callbacks.get(eventName);
     if (callbacks && callbacks.length > 0) {
       for (i = 0; i < callbacks.length; i++) {
-        callbacks[i].fn.call(callbacks[i].context || window, data);
+        callbacks[i].fn.call(callbacks[i].context || window, data, eventName);
       }
     } else if (this.failThrough) {
       this.failThrough(eventName, data);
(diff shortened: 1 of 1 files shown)


jpatel531:
Actually, might be easier if there was like a `patternBind` method like pawel said. Less magic involved

jpatel531:
yeah i'm gonna kill this. i overthought it

PR #153, opened 2016-03-24
Pattern matching on event names. Resolves #135

### LINKED ISSUE #135: [REQUEST] Event listener wildcards

jpatel531:
Allows you to pass an event pattern to a `bind_pattern` method.

```js
channel.bind_pattern(/events-/, function(data, match) {
  console.log("MESSAGE!!", data, match);
});
```

Rather than playing around with the plumbing, it just wraps around the `bind_all` method, like customers do already

cc @zimbatm @jackfranklin

### DIFF
--- a/src/events_dispatcher.js
+++ b/src/events_dispatcher.js
@@ -20,6 +20,19 @@
     return this;
   };
 
+  prototype.bind_pattern = function(pattern, userCallback){
+    if (!(pattern instanceof(RegExp))) {
+      Pusher.warn("::bind_pattern expects a RegExp");
+      return;
+    }
+    var callback = function(eventName, data){
+      if (pattern.test(eventName)) {
+        userCallback(data, eventName);
+      }
+    };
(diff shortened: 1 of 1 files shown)


jpatel531:
TODO:

[] Method to unbind from pattern
[] Tests

src/events_dispatcher.js
```diff
@@ -20,6 +20,19 @@
     return this;
   };
 
+  prototype.bind_pattern = function(pattern, userCallback){
+    if (!(pattern instanceof(RegExp))) {
+      Pusher.warn("::bind_pattern expects a RegExp");
+      return;
+    }
```
jackfranklin: people might expect to be able to pass a string in here which would be passed to `new RegExp` ? Wdyt?

Issue #148, opened 2016-03-13
Pusher.warn function duplicates each log statement


edwardotis:
Expected behavior: 
single log statement at warn level

Actual behavior:
single log statement at warn level
followed immediately by
single log statement using console.log

Example:

Pusher : Error : {"type":"WebSocketError","error":{"type":"PusherError","data":{"code":1006}}}
background.js:228
Pusher : Error : {"type":"WebSocketError","error":{"type":"PusherError","data":{"code":1006}}} background.js:228

 Pusher.warn = function() {
    var message = Pusher.Util.stringify.apply(this, arguments);
    if (window.console) {
      if (window.console.warn) {
        window.console.warn(message);
      } else if (window.console.log) {
        window.console.log(message);
      }
    }
    if (Pusher.log) {
      Pusher.log(message);
    }
  };

https://github.com/pusher/pusher-js/blob/567076130149a5ca88fc6deb19f0603247c62160/src/pusher.js#L116

jackfranklin:
Good spot - thank you!

hph:
Fixed in 0bd142fd630aa66ceb4dd0a622819f1bda473b9b.

PR #156, opened 2016-04-04
Make `warn` check for `Pusher.log` first

jackfranklin:
This fixes Issue #148 which notes that `Pusher.warn` logs twice if
you've got `Pusher.log` defined. Now we first check for `Pusher.log` and
use that before logging directly to console.

### DIFF
--- a/src/pusher.js
+++ b/src/pusher.js
@@ -115,15 +115,14 @@
 
   Pusher.warn = function() {
     var message = Pusher.Util.stringify.apply(this, arguments);
-    if (window.console) {
+    if (Pusher.log) {
+      Pusher.log(message);
+    } else if (window.console) {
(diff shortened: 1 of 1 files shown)


jackfranklin:
@jpatel531 what do you think?

jpatel531:
Will add this on the board for the new release

jackfranklin:
@jpatel531 did this make it in? I can't remember where we got to with this.

jpatel531:
Nah we didn't as we were still trying to figure out what to do if people supplied custom loggers that weren't just `console.log`

jackfranklin:
Ah yes - well let's close this anyway because it's going to be tough to rebase these changes, the issue is still open so we can come back to it. It's hardily urgent.

PR #161, opened 2016-05-02
Add support for wsPath config option [updated for 3.1]

rajivm:
Add support for wsPath config option (just like httpPath config option)

### DIFF
--- a/spec/javascripts/unit/isomorphic/transports/transports_spec.js
+++ b/spec/javascripts/unit/isomorphic/transports/transports_spec.js
@@ -28,6 +28,28 @@ describe("Transports", function() {
       );
     });
 
+    it("should generate correct unencrypted URLs with custom path prefix", function() {
+      var url = Transports.ws.hooks.urls.getInitial("foobar", {
+        encrypted: false,
+        hostUnencrypted: "example.com:123",
+        httpPath: "/path"
+      });
+      expect(url).toEqual(
+        "ws://example.com:123/path/app/foobar?protocol=7&client=js&version=" + VERSION + "&flash=false"
+      );
+    });
+ 
+    it("should generate correct encrypted URLs with custom path prefix", function() {
+      var url = Transports.ws.hooks.urls.getInitial("foobar", {
+        encrypted: true,
+        hostEncrypted: "example.org:321",
+        httpPath: "/path"
+      });
+      expect(url).toEqual(
+        "wss://example.org:321/path/app/foobar?protocol=7&client=js&version=" + VERSION + "&flash=false"
+      );
+    });
+
     it("should not have a resource file", function() {
       expect(Transports.ws.hooks.file).toBe(undefined);
     });
--- a/spec/javascripts/unit/web/transports/transports_spec.js
+++ b/spec/javascripts/unit/web/transports/transports_spec.js
@@ -37,6 +37,28 @@ describe("Transports", function() {
       );
     });
 
+    it("should generate correct unencrypted URLs with custom path prefix", function() {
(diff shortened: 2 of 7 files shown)


jpatel531:
hey @rajivm  - what is the use case for this?

rajivm:
@jpatel531 Unfortunately in some cases we have customers who need to deploy our app. in restricted on-premise environments which have no external internet access. In this case we are using poxa which emulates the Pusher API so we don't need to change our code base from our SaaS offering that uses Pusher.com. In this case, we want to be able to deploy everything in the on-premise environment under a single hostname to simplify the environment. The options in this case are either reserving /app and /apps for poxa, or properly name-spacing it under a path (which would require this change in the client).

Waffle, another customer of Pusher, has a similar setup: http://blog.waffle.io/how-to-build-for-github-enterprise/

rajivm:
Also -- I believe the travis failure is unrelated to this PR and seems to be from an issue fetching a dependency.

rajivm:
@jpatel531 Is this a change you will consider? I have updated my pull request for 3.1 / typescript.

zimbatm:
I think it's good to support that use-case but it's should be called something like `pathPrefix` and default to `/app`.

rajivm:
@zimbatm sure, that's a possibility -- I ended up changing it to httpPath to keep consistent w/ the hostEncrypted/httpPath etc. stuff used for the other protocol. Wouldn't the real-http case also be a `pathPrefix`?

Also -- in the pusher world today, aren't /apps and /app both paths off the root? So if I made `/app` a default `pathPrefix` then the `/apps` wouldn't use this prefix?

zimbatm:
Sorry I meant `/apps` would be the default prefix

hph:
This doesn't seem like a very common use case. Since this PR hasn't seen any activity in a long time, I'm inclined to close it. Are you relying on a custom build of pusher-js or is this no longer a requirement?

rajivm:
@hph We are still relying on a custom build for this parameter.

hph:
@rajivm Cool. I'll look into this after the weekend and we can see if we can get it merged soon. Sorry for the delay!

rajivm:
@hph Appreciate it!

hph:
@rajivm So I've taken a better look at this and it looks good, just a couple of things before merging:

- I agree with @zimbatm above that `pathPrefix` is better. I was originally going to suggest just `path`, but I think `pathPrefix` is more explicit.
- It would be nice to mention this new option in the Configuration section of the README - not absolutely required, though, as it's a rather uncommon use case.

Thanks!

rajivm:
@hph I have no problems changing it to `pathPrefix` -- however, this to me seems inconsistent with the rest of the configuration -- in the case of the sockjs/http protocol:
```
    httpHost: Defaults.sockjs_host,
    httpPort: Defaults.sockjs_http_port,
    httpsPort: Defaults.sockjs_https_port,
    httpPath: Defaults.sockjs_path,
    statsHost: Defaults.stats_host,
```

In this case, `httpPath` is the sockjs path prefix:
```
var path = (params.httpPath || "/pusher") + getGenericPath(key);
```

Therefore, I went with `wsPath` to remain consistent with this.

If I did `pathPrefix` then there is an open question of whether this applies to the other protocols (such as sockjs/http that already has its own prefix config with `httpPath`)

Let me know what you think.

hph:
@rajivm This will be included in the next release. Hopefully you still have use for it, despite the delay!

rajivm:
@hph That might be the most perfect timing! We were just working on upgrading our Pusher client when an engineer on the team was like this config option doesn't work -- and I was like let me look at my PR's status.

PR #171, opened 2016-05-16
Add a global unbind method. Closes #170

### LINKED ISSUE #170: Unbind not available
I have a connection that looks like so:
```javascript
self.io = new Pusher(self.pusher_api, {
        encrypted: self.pusher_encrypted,
        authEndpoint: self.pusher_auth_endpoint,
        authTransport: self.pusher_auth_transport,
      });
```

I can `bind` from `this.io` but cannot `unbind` an event. Looking at the prototype noticed that `unbind` is not available but under the `global_emitter`. 

coschedule

Is it possible to add an alias like bind to the Pusher class?

jackfranklin:


jackfranklin:
@jpatel531 what do you think?

jpatel531:
Looks good - can you call `make` and just git commit amend? It looks like you've just built the web and node outputs.

jackfranklin:
Will do. However merging this will then leave the version that's within the dist folder on GitHub different to the one on the Pusher CDN. How do we handle smaller changes like this?

jackfranklin:
@jpatel531 nudge, what's the latest?

PR #178, opened 2016-06-29
Move ts-loader from dependencies to devDependencies

shinnn:
I think this is a mistake in this commit 71a771cb85ba71f70f90ec302533b4db3ad66d9a.


jpatel531:
Ah, thanks!

shinnn:
@jpatel531 Can you publish a new version to npm? Currently we must install unused ts-loader whenever we install pusher-js from npm.

PR #179, opened 2016-07-07
Connection failure when app gets inactive causes a crash on iOS

jennysjogren:
The connection error being stringified by logger can be circular in
react native.


jpatel531:
ah excellent! thanks for bringing this up :)

jpatel531:
I'm going to try and make a 3.2 release as soon as I have the time with a few other changes - so this will be on npm then :)

jennysjogren:
Sounds good! thank you :)

PR #180, opened 2016-07-12
edited readme

akram-rameez:
Just changed a line in readme. the include script wasnt correct.


jpatel531:
Aha, this might more be a problem with forgetting to publish to bower. Ill check that when im in the office. Pusher js iso is a deprecated library that was the basis for pusher js 3.1

PR #181, opened 2016-07-12
Update README.markdown

sebas932:
Updated Bower documentation according to the install.


hph:
@sebas932 The configuration was actually wrong, so the readme does not need to be updated. This has been fixed in 4.0.0.

PR #187, opened 2016-07-29
Version 3.2

### LINKED ISSUE #168: CSP issue with 3.1 in Chrome extension background context
Ran into an issue with the latest release today. In a chrome extension context eval of any form is not allowed by default. This causes a problem with how you find the global object.

The offending line of code is [here](https://github.com/pusher/pusher-js/blob/7d4160038b5836163af2bdc23aabce6219a9c242/src/core/base64.ts#L1)

```
var global = Function("return this")();
```

There is a way to relax this constraint but I would rather not expose my extension to XSS attacks. For now 3.0 works fine for my needs, but it would be nice to get updates in the future. Would it be possible to replace these assignments with your webpack build process for each environment? So the web version would have `var global = window;` and `var global = self;` in web worker etc.

jpatel531:
* Kill get global hack in favour of define plugin. Resolves #168
* Use Douglas Crockford's decycle as a fallback to failed JSON.stringify calls

@jackfranklin ?


jackfranklin:
LGTM but probably worth making sure travis is happy. It looks like all the tests ran but something timed out.

jpatel531:
Mm. Maybe I should get rid of the travis github integration and add a note to anyone planning to release the lib to run automated browser testing locally

jpatel531:
I was so excited about CI too! :(

PR #191, opened 2016-09-01
Make it clear that disableStats do not affect the user account stats

luismfonseca:
Just to be clear.

/cc @jpatel531


hph:
@luismfonseca Thank you for the contribution 🎁

PR #196, opened 2016-11-01
Prod 1187 subscribe unsubscribe bugfix

hot-leaf-juice:
This is a proposed fix for PROD-1187.

# Problem
Unsubscribing from a private or presence channel very soon after subscribing results in an inconsistent state where the channel is no longer listed in the Pusher object, but you continue to receive events on it. This situation could conceivably arise in the wild when a user triggers a subscribe and an unsubscribe event in quick succession.

# Steps to reproduce
```
pusher.subscribe('private-ch'); pusher.unsubscribe('private-ch');
```
Then send an event on `private-ch` and it will be received.

# Diagnosis
The external unsubscribe call throws an error since it is trying to unsubscribe from a channel that hasn't actually been subscribed to yet, however the reference to the channel is successfully removed from the Pusher object. After this the subscription completes and we receive events despite not having a reference to the channel.

# Solution
Track whether channels are pending subscription. If we attempt to unsubscribe from a channel that is still pending subscription, then rather than unsubscribing immediately, toggle a `subscriptionCancelled` flag instead. Then unsubscribe properly upon receiving `pusher_internal:subscription_succeeded`, while also suppressing the `pusher:subscription_succeeded` event so that from the user's perspective, the subscription was cancelled mid process.


spec/javascripts/unit/core/channels/presence_channel_spec.js
```diff
@@ -156,6 +164,43 @@ describe("PresenceChannel", function() {
...
+          });
+          channel.handleEvent("pusher_internal:subscription_succeeded");
+        });
+
+        it("should set #subscriptionPending to false", function() {
+          channel.cancelSubscription();
+          channel.bind(function() {
+            expect(channel.subscriptionPending).toEqual(true);
+          });
+          channel.handleEvent("pusher_internal:subscription_succeeded");
```
hot-leaf-juice: I copied the format of these from one of the existing tests and on closer inspection I'm not really sure what they are doing. I'm not sure what binding without specifying an event even does... will rewrite these so they make more sense to me.

spec/javascripts/unit/core/pusher_spec.js
```diff
@@ -310,13 +310,31 @@ describe("Pusher", function() {
...
+        var channel = pusher.subscribe("yyy");
+        expect(pusher.channel("yyy")).toBe(channel);
+
+        pusher.unsubscribe("yyy");
+        expect(pusher.channel("yyy")).toBe(undefined);
+      });
+
+      it("should delay unsubscription if the subscription is pending", function () {
+        var channel = pusher.subscribe("yyy");
+        channel.subscriptionPending = true;
```
hot-leaf-juice: The channel object in the pusher_spec tests is just a mock, so it doesn't do anything except track when it has been called :) The real channel object does indeed set `subscriptionPending` when you call `.subscribe`. This test just makes sure that when the subscription is still pending, the `pusher` object doesn't try to call `.unsubscribe()` on the channel, and instead calls `.cancelSubscripiton()`.

spec/javascripts/unit/core/channels/private_channel_spec.js
```diff
@@ -108,6 +116,43 @@ describe("PrivateChannel", function() {
...
+
+        expect(callback).not.toHaveBeenCalled();
+      });
+
+      it("should set #subscribed to true", function() {
+        channel.cancelSubscription();
+        channel.bind(function() {
+          expect(channel.subscribed).toEqual(true);
+        });
+        channel.handleEvent("pusher_internal:subscription_succeeded");
```
hot-leaf-juice: likewise ^

spec/javascripts/unit/core/channels/private_channel_spec.js
```diff
@@ -108,6 +116,43 @@ describe("PrivateChannel", function() {
...
+      });
+
+      it("should set #subscriptionPending to false", function() {
+        channel.cancelSubscription();
+        channel.bind(function() {
+          expect(channel.subscriptionPending).toEqual(true);
+        });
+        channel.handleEvent("pusher_internal:subscription_succeeded");
```
hot-leaf-juice: likewise ^

hot-leaf-juice:
Updated with changes 👍 Only touched the tests.

brettjonesdev:
Hey, sorry, late to the party here... I initially reported this bug, so I just wanted to check in and see about one more test case that would be good to nail down.  Does this PR handle the following scenario?  

```
channel.subscribe('private-ch');
channel.unsubscribe('private-ch');
channel.subscribe('private-ch');
// send an event on private-ch, it should be received
```

Just checking that this solution works not only for the subscribe/unsubscribe, but for any number of chained subscribe/unsubscribe calls.  I'm picturing a scenario where a UI component gets initialized, torn down and recreated synchronously - want to make sure that the _last_ operation, whether subscribe or unsubscribe, is applied properly.  Thoughts?

hot-leaf-juice:
eek. I'm afraid by the looks of it this actually won't work if the third statement is executed while the original subscription is still pending. Should be a pretty simple fix though, just also need to account for cancelling a cancellation! Will get another fix pushed asap.

– Callum

brettjonesdev:
Haha okay, awesome! (Didn't really hot-leaf-juice was you, Callum ;) )

And you _know_ I'm going to follow up with `sub/unsub/sub/unsub/sub` 😆

hot-leaf-juice:
Of course! We'll try and handle arbitrary chains this time. I'm slightly embarrassed that I didn't think of this first time around...

brettjonesdev:
Haha, no worries. I only thought of it cuz of my recent experience with my
own realtime code which _didn't_.  Wound up with an absurdly convoluted
solution where I chained up promises for a given channel.  :P

hot-leaf-juice:
@brettjonesdev see #201 for progress on this.

Issue #62, opened 2013-12-10
Authentication should support CORS


ryansmith3136:
I am going to be brief with my bug report. If you need a more thorough description, let me know.

I am assuming that to properly handle the client's request to `/pusher/auth` I need to use cookies.

I expected that I could have pusher-js authenticate against my backend HTTP service across origins. I realized that the requests to `/pusher/auth` did not contain any cookies. Upon further inspection, I found that the pusher-js makes a new XHR object to send to the authentication endpoint of my service. This XHR object does not touch the `withCredentials` field which, I believe, is required for CORS.

My request is that you provide support for authenticating across origins. Perhaps that is by setting the `withCredentials` field.

chris-martin:
I just spent a while tracking this down. The documentation perhaps could make it a little more clear that jsonp is the *only* supported way to do cross-origin auth.

fiatjaf:
Today you would have to do

```javascript
Pusher.Runtime.createXHR = function () {
  var xhr = new XMLHttpRequest()
  xhr.withCredentials = true
  return xhr
}
```

tadas-s:
Hello,

Are there any plans to implement this? 4 years later, IE9 is largely irrelevant browser.

Cheers guys

PR #198, opened 2016-11-03
WIP Xhr with credentials

jpatel531:


hph:
Hi @jpatel531! Thanks for opening this PR. I was just wondering whether you had any concrete plans or vision for this or where it should go, since it seems to be a work in progress (also with conflicts and no tests). Thoughts?

jpatel531:
ah - i did this on my penultimate day! i think it was for a support ticket and issues like #62 . no vision really, just wanted to fix it for a guy on support. it could probably be made better with a good custom authorizer abstraction. i'm not a fan of how the fetch, json and xhr authorizer work  - probably could be cleaner, but at the time i tried to limit changing the internals too much and just separating out a function seemed the easiest route.

hph:
Okay, thank you for the input!

PR #201, opened 2016-11-07
Subscription unsubscription robustification

hot-leaf-juice:
Following on from the discussion in #196, adds a method to reinstate a cancelled subscription. We should now be able to handle arbitrary chains of subscriptions and unsubscriptions – where the final method is always respected – regardless of the timings.

e.g. subscribe -> unsubscribe -> subscribe -> subscribe -> unsubscribe -> subscribe = subscribe

@hph

### DIFF
--- a/spec/javascripts/unit/core/channels/channel_spec.js
+++ b/spec/javascripts/unit/core/channels/channel_spec.js
@@ -218,7 +218,7 @@ describe("Channel", function() {
       channel.subscribe();
 
       expect(channel.subscriptionPending).toEqual(false);
-    }
+    });
   });
 
   describe("#unsubscribe", function() {
--- a/spec/javascripts/unit/core/channels/channel_spec.js
+++ b/spec/javascripts/unit/core/channels/channel_spec.js
@@ -241,4 +241,15 @@ describe("Channel", function() {
       expect(channel.subscriptionCancelled).toEqual(true);
     });
   });
+
+  describe("#reinstateSubscription", function() {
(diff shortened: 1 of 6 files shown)


brettjonesdev:
Those specs are glorious.  👍

spec/javascripts/unit/core/channels/channel_spec.js
```diff
@@ -204,13 +204,21 @@ describe("Channel", function() {
...
       expect(channel.subscriptionPending).toEqual(true);
     });
+
+    it("should do nothing if already subscribed", function() {
+      channel.subscribed = true;
+
+      channel.subscribe();
+
+      expect(channel.subscriptionPending).toEqual(false);
+    }
```
hph: This should be `});`, it's a syntax error as is.

hph: I'll fix this myself.

hot-leaf-juice: eek. will do. Not sure what's going on with that syntax error. I'm sure that should have been thrown up by the unit tests. I might have merged something sloppily.

hot-leaf-juice: They run on my end? Note that they are *integration* tests not unit. (since you need a fully working pusher object with real non mocked channels to test the interaction).

hph: My mistake. Again, great work, and thanks!

PR #202, opened 2016-11-22
Prod 1211 unbind all

hot-leaf-juice:
A customer pointed out that `unbind_all` didn't do what they expected. In particular it didn't reverse `bind_all`. Upon further inspection `unbind_all` was implemented identically to `unbind` and in fact all occurrences of `unbind_all()` can safely be replaced by `unbind()`. This also means that there is no way of removing global bindings (of the `bind_all` variety) short of directly mutating `channel.global_callbacks`.

After discussion in Slack we came to the conclusion that an `unbind_all` method should do just that, remove *all* bindings, and that `bind_all` should be replaced by `bind_global` and it's inverse `unbind_global` to avoid ambiguity.

This is a breaking change, but any use of `unbind_all()` will continue to work as expected (unless anyone was exploiting the fact that `unbind_all()` previously left global bindings alone) and `bind_all` seems to have been discouraged in the docs, so the change to `bind_global` shouldn't affect a huge number of people.

### DIFF
--- a/src/core/events/dispatcher.ts
+++ b/src/core/events/dispatcher.ts
@@ -1,5 +1,6 @@
-import CallbackRegistry from './callback_registry';
+import * as Collections from '../utils/collections';
 import Callback from './callback';
+import CallbackRegistry from './callback_registry';
 
 /** Manages callback bindings and event emitting.
  *
@@ -21,18 +22,33 @@ export default class Dispatcher {
     return this;
   }
 
-  bind_all(callback : Function) {
+  bind_global(callback : Function) {
     this.global_callbacks.push(callback);
     return this;
   }
 
-  unbind(eventName : string, callback : Function, context?: any) {
+  unbind(eventName? : string, callback? : Function, context?: any) {
     this.callbacks.remove(eventName, callback, context);
     return this;
   }
 
-  unbind_all(eventName?: string, callback?: Function) {
-    this.callbacks.remove(eventName, callback);
+  unbind_global(callback?: Function) {
+    if (!callback) {
+      this.global_callbacks = [];
+      return this;
+    }
+
+    this.global_callbacks = Collections.filter(
+      this.global_callbacks || [],
+      c => c !== callback
+    );
+
+    return this;
+  }
+
+  unbind_all() {
+    this.unbind();
+    this.unbind_global();
     return this;
   }
 
--- a/src/core/pusher.ts
+++ b/src/core/pusher.ts
@@ -183,12 +183,12 @@ export default class Pusher {
     }
   }
 
-  bind(event_name : string, callback : Function) : Pusher {
-    this.global_emitter.bind(event_name, callback);
+  bind(event_name : string, callback : Function, context? : any) : Pusher {
+    this.global_emitter.bind(event_name, callback, context);
     return this;
   }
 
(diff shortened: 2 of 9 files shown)


src/core/pusher.ts
```diff
@@ -180,18 +180,28 @@ export default class Pusher {
...
 
-  bind(event_name : string, callback : Function) : Pusher {
-    this.global_emitter.bind(event_name, callback);
+  bind(event_name : string, callback : Function, context? : any) : Pusher {
+    this.global_emitter.bind(event_name, callback, context);
     return this;
   }
 
-  unbind(event_name: string, callback: Function) : Pusher {
+  unbind(event_name? : string, callback? : Function, context? : any) : Pusher {
```
hph: `context` should probably not be passed here, unless you were going to do something with it. The two `unbind` functions appear to be inconsistent since one handles context and the other does not (before and in this PR).

src/core/events/dispatcher.ts
```diff
@@ -21,18 +22,33 @@ export default class Dispatcher {
...
   }
 
-  bind_all(callback : Function) {
+  bind_global(callback : Function) {
     this.global_callbacks.push(callback);
     return this;
   }
 
-  unbind(eventName : string, callback : Function, context?: any) {
+  unbind(eventName? : string, callback? : Function, context?: any) {
```
hph: This API is a bit dangerous, since arguments passed to the function could resolve to undefined (to due a bug in a client's use of the library) and then that would result in unbinding all listeners. However, that should not be part of this PR.

src/core/pusher.ts
hot-leaf-juice: I'm not sure I follow. All pusher's `unbind` function does is delegate to the `unbind` function of `EventsDispatcher`? I'm just allowing you to set the `context` in the same way you already can on the channel.

hot-leaf-juice: https://github.com/pusher/pusher-js/blob/master/src/core/events/dispatcher.ts#L29

src/core/events/dispatcher.ts
hot-leaf-juice: According to the existing documentation and the underlying function, being able to call `unbind("event-name")` or `unbind()` is an intended use, so I just thought the types of the parameters in the typescript should respect that. It doesn't make any changes in the js! Totally agreed on it being kinda dangerous though. Would be a much bigger change to change that behaviour.

mdpye:
I'm not sure what the process is for change logging in pusher-js, but I'd suggest adding a `next release` section to the changelog and filling out the details in this PR, so that it can't be forgotten, and the person running the release doesn't need to research the API alterations again to put together a changelog entry.

hph:
@mdpye Sounds like something perfect for a PR_TEMPLATE! Good idea.

hot-leaf-juice:
good plan

src/core/pusher.ts
hph: What I meant is that your change added the parameter to the function but the parameter itself remains unused. I assume that you intended to pass it to `this.global_emitter.unbind`, which after taking a look into the Dispatcher appears correct.

hot-leaf-juice: Oooooooh! Yes, that is exactly what I meant. :) Sorry!

Issue #165, opened 2016-05-06
Configurable authentication


sandstrom:
If the pusher js-client could take a function `authAdapter` returning a promise, then it could be left to the user to perform the authentication in any way fit (different endpoints, headers handling, withCredentials switch, etc).

Should solve https://github.com/pusher/pusher-js/issues/83 and https://github.com/pusher/pusher-js/issues/99.

``` js
new Pusher(apiKey, {
  authAdapter: function(channel, ...) {
    return Promise(resolve, reject) {
      // ...
    };
  },
  // other conf...
});
```

jpatel531:
I like the idea. I might be sceptical about Promises as we'd have to add another dependency to polyfill older browsers. Auth adapters is definitely something we'd like to tackle after the coming release is out

sandstrom:
It has wide support (http://caniuse.com/#search=promise) and the shim is small: 
https://github.com/stefanpenner/es6-promise (3.4 kb gzip + minified)

Entirely up to you of course. Callbacks works too!

jackfranklin:
Although I'd personally prefer promises, I would stick to callbacks given our commitment to supporting a bunch of older browsers as well as newer ones.

Or we could make it work with both and ask those who want to use Promises to provide their own polyfill.

Like the idea though 👍

sandstrom:
@jpatel531 friendly ping

hph:
I'm closing "feature request" issues, including this one. This does not mean that the feature will not be implemented, we just want to reserve open GitHub issues to bugs that need to be addressed.

Feel free to comment here to keep the discussion going. People watching the repository will still receive notifications.

sandstrom:
AFAIK this was implemented (contrary to hph's message above): #211

thecodecafe:
Hi, so I'm setting up pusher but can't seem to find a way to define the domain for authentication, I'm using pusher on a  react-native app, and what to create a private channel but need the user authenticated before creating a private channel, below is my config code.

```
let pusher new Pusher(
        PUSHER_KEY, {
            cluster: 'eu',
            forceTLS: true,
            authEndpoint: `/pusher/auth`,
            auth: {
                headers: {
                    Authorization: `Bearer ${user && user.token}`
                }
            }
        }
    )
```

how can I define a domain for that endpoint, have tried adding the domain to the endpoint but that didn't work, also I'm not so sure if the request is being made from the device or from Pusher's server?

Really stuck here, thanks.

[...]

PR #211, opened 2017-01-17
Allow specifying a custom authorizer

ejlangev:
## What does this PR do?

This PR adds an option to use when configuring the Pusher client to override the behavior of the authorization.  My original use case was needing to pass a header along with my request that was only known at the time of the request and this solves that problem but also seems to solve a few other open issues.  Most importantly https://github.com/pusher/pusher-js/issues/165.

Open to suggestions for improvements.

## Checklist

- [x] All tests are passing.

/cc @hph

### DIFF
--- a/README.markdown
+++ b/README.markdown
@@ -156,6 +156,23 @@ var pusher = new Pusher(API_KEY, {
 });
 ```
 
+#### `authorizer` (Function)
+
+If you need custom authorization behavior you can provide your own `authorizer` function as follows:
+
+```js
+var pusher = new Pusher(API_KEY, {
+  authorizer: function(channel, options) {
+    return {
+      authorize: function(socketId, callback) {
+        // Do some ajax to get the auth information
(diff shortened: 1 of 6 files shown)


hph:
Thanks, @ejlangev. We'll take a look at this shortly.

duxet:
Without custom authorizer it's not possible to subscribe Travis private channels :(

hot-leaf-juice:
@duxet you can already add custom authorisers so you don't need to wait for this PR to be accepted. See [here](https://github.com/pusher/pre-auth-example/blob/master/script.js) for an example. From a brief look, it seems that this PR just makes this process much easier (which is great! because the current way is a little bit of a hack :) )

ejlangev:
@hph Anything else I need to do for this?

hph:
Hi @ejlangev! Sorry for the delay, this looks good to me!

Issue #158, opened 2016-04-25
Avoid Global Namespace


thetutlage:
I developed an embeddable widget which makes use of Pusher. Pusher is required from [npm](https://www.npmjs.com/package/pusher-js) and my final js file is bundled using `browserify`.

As Pusher makes use of `window.Pusher` it is creating conflicts when the page using my widget is already making use of Pusher.

jackfranklin:
We should probably only export onto the window if there's nothing there already.

@jpatel531 what do you think?

thetutlage:
But when if the global Pusher version is something different from the one bundled using require js.

So in my example the 3rd party website was using pusher `2.2` and i built my widget using `3.0`. When my widget loaded it added `3.0` to the global scope and their Pusher code failed as they are dependent upon something from `2.2`.

jackfranklin:
There's only so much we can do here; we need Pusher to be globally available because a lot of people use us in that manner. What I've done in my own code is make my widgets take an instance of Pusher as an argument, and then you don't need your widget to bundle Pusher.

I also would encourage you to upgrade to v3 too as it's a fairly large improvement.

Additionally, the issues of managing different versions of Pusher on the same page isn't something we're able to easily fix.

thetutlage:
Okay, make sense. Thanks for quick response

jpatel531:
+1 @jackfranklin

mgalgs:
I just ran into this issue due to [the Freshdesk Live Chat widget](https://freshdesk.com/multichannel-support/live-chat) bundling its own copy of Pusher.

> we need Pusher to be globally available because a lot of people use us in that manner.

Yes, you don't want to break backwards compatibility, but couldn't there be an optional "noconflict" mode?  Something like [`jQuery.noConflict()`](https://api.jquery.com/jquery.noconflict/).

> What I've done in my own code is make my widgets take an instance of Pusher as an argument, and then you don't need your widget to bundle Pusher.

But what if you don't control the widget (as is the case for me with the Freshdesk widget)?

hot-leaf-juice:
I haven't tested this myself yet so ymmv, but you might have some luck bundling the node build of pusher-js with webpack or something similar, and then importing it under some other name. e.g. something like

    let Pusher2 = require("pusher-js/node");

As far as I'm aware the "node" version doesn't actually rely on any runtimes, and the main difference is just the module definition format, so the above should work in the browser.

mgalgs:
[...]

PR #213, opened 2017-01-28
worker: fetch_auth: Send credentials

mgalgs:
## What does this PR do?

By default the Fetch API doesn't send any cookies, even for same-origin
requests [1].  This presents a problem when using the fetch API against a
web framework that relies on cookies for authentication (like Django).

Note that the worker runtime is typically not used from the browser, except
when needed to work around issues with conflicts between multiple versions
of Pusher existing on the same page.  One example of when this happens is
if a third-party widget is bundling Pusher, and the main app is also trying
to use it.  For example: [2].

Without this patch, it's not possible to set up authenticated connections
against a Django app with CSRF protection enabled while using the worker
runtime.  With the patch, all's well.

[1] https://developer.mozilla.org/en-US/docs/Web/API/Request/credentials
[2] https://github.com/pusher/pusher-js/issues/158#issuecomment-274242084

## Checklist

- [X] All new functionality has tests. (no new functionality (depending on how you defined "functionality" ;) )
- [X] All tests are passing. (Ran for about 5 minutes and seemed to be looping endlessly?)
- [X] New or changed API methods have been documented. (N/A)

/cc @hph @hot-leaf-juice

### DIFF
--- a/src/runtimes/worker/auth/fetch_auth.ts
+++ b/src/runtimes/worker/auth/fetch_auth.ts
@@ -14,6 +14,7 @@ var fetchAuth : AuthTransport = function(context, socketId, callback){
   var request = new Request(this.options.authEndpoint, {
     headers,
     body,
+    credentials: "same-origin",
     method: "POST",
   });
 


hph:
Hi @mgalgs and thanks for the contribution! I think this looks okay. If there is a need for it, we can later provide an option to set `credentials` to `omit`, `same-origin` or `include`.

PR #224, opened 2017-04-13
Change .done() to .then() in react-native

kubik369:
## What does this PR do?

Calling .done() crashed app on react-native@0.42.3 with .done undefined.

## Checklist

- [ ] All tests are passing.

/cc @hph

### DIFF
--- a/src/runtimes/react-native/net_info.ts
+++ b/src/runtimes/react-native/net_info.ts
@@ -14,7 +14,7 @@ export class NetInfo extends EventsDispatcher implements Reachability {
     super();
     this.online = true;
 
-    NativeNetInfo.fetch().done((connectionState)=>{
+    NativeNetInfo.fetch().then((connectionState)=>{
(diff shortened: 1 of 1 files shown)


hph:
This is weird. Did React Native use a different implementation of Promise at some point? In any case, this seems reasonable, I'm just wondering whether we need to support older versions of React Native as well.

kubik369:
According to most docs, `.done` is documented everywhere only due to historical reasons and also non-standard (`https://www.promisejs.org/api/`), so `.then` should work just fine even on older versions.

hph:
Okay, thanks. I'll merge this and include it in the next version.

kubik369:
Do you have any estimates as to when will you release the next version as we would like to use your library without the need to run `sed` after each colleague installs the package.

hph:
@kubik369 I expect to release a new version on Tuesday.

PR #240, opened 2017-07-06
More helpful warnings with links to docs

leesio:
## What does this PR do?
Adds some links to relevant pusher docs when warnings are logged. Included a `UrlStore` for future expansion/easy changes.

## Checklist

- [X] All new functionality has tests.
- [X] All tests are passing.
- [-] New or changed API methods have been documented.

/cc @hph


src/runtimes/isomorphic/auth/xhr_auth.ts
```diff
@@ -34,7 +35,11 @@ var ajax : AuthTransport = function(context : AbstractRuntime, socketId, callbac
           callback(false, data);
         }
       } else {
-        Logger.warn("Couldn't get auth info from your webapp", xhr.status);
+        var suffix = UrlStore.buildLogSuffix("authentication_endpoint");
+        Logger.warn(
+          `Couldn't retrieve authentication info. ${status}` +
```
hph: Shouldn't this be `xhr.status`? You might also need to add a space after it.

src/core/utils/url_store.ts
```diff
@@ -0,0 +1,38 @@
...
+*
+* @param {string} key - relevant key in the url_store.urls object
+* @return {string} suffix string to append to log message
+*/
+const buildLogSuffix = function(key: string) : string {
+  const url_prefix = "Check out:";
+  const url_obj = url_store.urls[key];
+  if(!url_obj) return "";
+
+  var url;
```
hph: `let` instead of `var`.

src/core/utils/url_store.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (url_obj.full_url) {
+    url = url_obj.full_url;
+  } else if(url_obj.path) {
+    url = url_store.base_url + url_obj.path;
+  }
+
+  if(!url) return "";
+  return [url_prefix, url].join(" ");
```
hph: I think I'd prefer using template strings consistently, since you introduce them elsewhere. So in this case this could just become \`${urlPrefix} ${url}\`, although in a way I kind of like `[].join(' ')`, too.

src/core/utils/url_store.ts
```diff
@@ -0,0 +1,38 @@
...
+}
+
+/** Builds a consistent string with links to pusher documentation
+*
+* @param {string} key - relevant key in the url_store.urls object
+* @return {string} suffix string to append to log message
+*/
+const buildLogSuffix = function(key: string) : string {
+  const url_prefix = "Check out:";
+  const url_obj = url_store.urls[key];
```
hph: I think we should try to enforce consistency by using camelcase everywhere in new code, even though there are parts that are currently inconsistent.

src/core/channels/presence_channel.ts
```diff
@@ -25,10 +26,12 @@ export default class PresenceChannel extends PrivateChannel {
     super.authorize(socketId, (error, authData) => {
       if (!error) {
         if (authData.channel_data === undefined) {
+          let suffix = UrlStore.buildLogSuffix("authentication_endpoint");
           Logger.warn(
             "Invalid auth response for channel '" +
             this.name +
-            "', expected 'channel_data' field"
+            "', expected 'channel_data' field." +
+            suffix
```
hph: You could use a template string here, too.

src/runtimes/isomorphic/auth/xhr_auth.ts
leesio: Fixed in c562077

src/core/channels/presence_channel.ts
leesio: Fixed in c562077

src/core/utils/url_store.ts
leesio: Fixed in c562077

src/core/utils/url_store.ts
leesio: Fixed in c562077

src/core/utils/url_store.ts
leesio: Fixed in c562077

leesio:
Addressed all above in c562077. Also changed filename to `-` instead of `_`. I now think there are no underscores at all in this PR 👍

hph:
@leesio Actually, I think most (if not all) files with multiple words in them user underscores, so we should probably stick to that. I prefer dashes in filenames, but it doesn't really make a difference.

leesio:
Ha! Updated

hph:
"A foolish consistency is the hobgoblin of little minds", and all that, but oh well.

src/core/pusher.ts
```diff
@@ -67,6 +68,12 @@ export default class Pusher {
   constructor(app_key : string, options : any) {
     checkAppKey(app_key);
     options = options || {};
+    if (!options.cluster) {
+      let suffix = UrlStore.buildLogSuffix("javascript_quick_start");
+      Logger.warn(
+        `You should always specify a cluster when connecting. ${suffix}`
```
shinnn: However, the [Pusher document](https://pusher.com/docs/clusters) says:

> The default cluster, in us-east-1, has a shortcode of `mt1`. You currently don’t need to specify this as it’s assumed as the default, but there’s no harm in doing so

Actually, though my application uses the default `mt1` cluster, Pusher.js logs a bit strict message "You should always ...".

@hph May I create a pull request to remove this warning? Or, the Pusher document is misdirecting the appropriate usage?

hph: Hey @shinnn, the documentation is out of date so I'll update the docs.

shinnn: Long time no see, @hph. 

> the documentation is out of date so I'll update the docs.

The [doc](https://pusher.com/docs/clusters) doesn't seem to be updated and still says:

> You currently don’t need to specify this as it’s assumed as the default, but there’s no harm in doing so.

Can you tell me how will you update it in the future? Will you remove the "`mt1` by default" statement?

shinnn: https://github.com/pusher/pusher-js/pull/261 was merged, that proves my first comment was right.

hph: Hi @shinnn! The documentation changes were merged today but they haven't been deployed yet. I was waiting on that before replying to you, but now you know :) We definitely don't want to have an implicit default cluster, so there's no need for a PR to change it, but thank you for offering!

PR #247, opened 2017-08-24
use ws to replace faye-websocket

pluswave:
## What does this PR do?

For wss support in nodejs, ws package works but faye-websocket doesn't. and ws package seems more stable.

/cc @hph


hph:
Hi @pluswave, and thank you for the PR.

Even though this library works in Node.js, you should use [pusher-http-node](https://github.com/pusher/pusher-http-node) instead. As for these changes, to me it looks like faye-websocket does in fact support encrypted connections, although I might be wrong. In any case, I don't see a good reason for switching to `ws` from faye-websocket, but if we did want to switch libraries then [µWS](https://github.com/uNetworking/uWebSockets) is probably a better choice.

Again, thank you for the time and effort spent opening the PR. However, since these changes don't seem to bring any meaningful new features or benefits to the table I'm going to close this PR. If there's something I'm missing, please don't hesitate to get in touch and we can re-evaluate whether doing this makes sense.

Issue #223, opened 2017-04-10
Pusher doesn't receive messages after a network change


vkulov:
I'm having issues with pusher sometimes not receiving messages on a channel it is subscribed to. It happens randomly for our clients. We managed to reproduce it consistently by changing the network adapter we use for internet connectivity. The test was conducted on both Windows 10 and a Mac OS X machine under Chrome browser. We use your test page on https://test.pusher.com

First we open the test page and verify that both Ajax and Client triggers work properly.

![image](https://cloud.githubusercontent.com/assets/134308/24883819/4e3cee4e-1e4f-11e7-82b6-9543e38f2a40.png)

Then we disconnect the network by unplugging the cable connection. Pusher detects that there's no connection.

![image](https://cloud.githubusercontent.com/assets/134308/24883862/77c69d1e-1e4f-11e7-9b27-bfd8589f1f31.png)

Then we connect to a wi-fi network backed by a mobile hotspot.

![image](https://cloud.githubusercontent.com/assets/134308/24883920/c5f1665e-1e4f-11e7-98f1-b4f29af5925c.png)

Pusher successfully re-connects and we verify that it works. We try this one more time. Going back and forth between the cable connection and the wi-fi mobile hotspot (until we get the 1006 error message.)

![image](https://cloud.githubusercontent.com/assets/134308/24883978/002511fe-1e50-11e7-923c-9c3e87d6b66a.png)

After the network is reconnected pusher discovers that and says 'connected'. However the Ajax Trigger button doesn't print anything in the log. And clicking the Client Trigger shows an error message "Cannot broadcast client event (connection not subscribed to channel presence-channel)"

![image](https://cloud.githubusercontent.com/assets/134308/24884037/4abbb6d2-1e50-11e7-8649-ed2f2d701b47.png)

Clicking Disconnect and then Connect buttons doesn't recover Pusher. It prints out [status] disconnected and then [status] connected but the Ajax Trigger button still doesn't work and the Client Trigger button still prints the same error.

We verify in Network tab on chrome that the Ajax Trigger button successfully sends the 'hello' request to pusher server and received a response from server 'ok' but the we get nothing printed in the logs.

We are able to reproduce this consistently. Let me know if we can get any more debug info for you.

hph:
@vkulov Thanks for reporting this issue and describing the steps to reproduce it in detail. I'm looking into the matter and I will get back to you soon.

andefred:
[...]

PR #251, opened 2017-08-31
reset subscriptionPending if disconnect happens while subscribing

hengwoon:
## What does this PR do?

If Pusher disconnects (manually, or when network blips out and changes) while a channel subscription is happening (but is not yet successful), the channel is stuck in a pending subscription state. This means that when Pusher re-establishes connection and tries to re-subscribe channels, it skips over channels that are pending subscription. 

This results in the issue seen in #223 , and can be reproduced by going to http://test.pusher.com/, click disconnect button, then quickly click the connect and then disconnect again before the pusher channel subscription succeeds.

This fix resets `Channel.subscriptionPending` to `false` when a disconnect happens.

## Checklist

- [x] All tests are passing.

/cc @hph

### DIFF
--- a/spec/javascripts/unit/core/channels/channel_spec.js
+++ b/spec/javascripts/unit/core/channels/channel_spec.js
@@ -64,6 +64,14 @@ describe("Channel", function() {
       channel.disconnect();
       expect(channel.subscribed).toEqual(false);
     });
+
+    it("should set subscriptionPending to false", function() {
+      channel.subscriptionPending = true;
+      
+      channel.disconnect();
+
+      expect(channel.subscriptionPending).toEqual(false);
+    });
   });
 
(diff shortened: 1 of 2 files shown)


hengwoon:
I left out the build files to keep this clean, plus my build files ended up changing a lot more than necessary (snake case to camel case) for some reason.

hph:
Great work, @hengwoon! It looks like this does the trick. Don't worry about the built files, I can generate them before the release.

tommoor:
Could we get an updated release on npm with this fix? 😄  The last version there is 5 months old…

hph:
Sorry, @tommoor - somehow this got lost in my notifications. Version 4.2.1 was released on NPM a few days back, check it out!

PR #278, opened 2018-04-22
consistent naming of socket var

jameshfisher:
## What does this PR do?

Use `const socket` everywhere

/cc @hph


jameshfisher:
unfortunately it's inconsistent with the dashboard, which uses `pusher = new Pusher` everywhere, but that's something to tackle another time ...

PR #287, opened 2018-07-09
End to end encryption

leesio:
## What does this PR do?

[Description here]

## Checklist

- [x] All tests are passing. (Web and node)

/cc @hph


src/core/channels/channels.ts
```diff
@@ -62,6 +62,13 @@ export default class Channels {
 
 function createChannel(name : string, pusher : Pusher) : Channel {
   if (name.indexOf('private-encrypted-') === 0) {
+    // We don't currently support e2e on React Native due to missing functionality.
+    // This prevents any weirdness by just returning a private channel instead.
+    if(navigator.product == "ReactNative") {
+      let errorMsg = `Encrypted channels are not yet supported when using React Native builds.`;
+      Logger.warn(`Error: ${errorMsg}`);
```
WillSewell: Why not just throw an exception? Carrying on with something non-obvious is more likely to lead to confusion IMO (not looking at the debug console/logger not be configured properly/"why are my events not being decrypted??")

kn100: @leesio thoughts?

PR #291, opened 2018-07-31
Remove support for client events on encrypted channels

WillSewell:
## What does this PR do?

In title.

We may decide to support client events in the future, in which case this commit can be reverted.

## Checklist

- [ ] All tests are passing.


src/core/channels/encrypted_channel.ts
```diff
@@ -41,10 +41,10 @@ export default class EncryptedChannel extends PrivateChannel {
     });
   }
 
-  trigger(event: string, data: any) {
-    if(!this.key) return false;
-    let encryptedData = this.encryptPayload(data);
-    return super.trigger(event, encryptedData);
+  trigger(event: string, data: any): boolean {
+    throw new Errors.UnsupportedFeature(
+      'Client events are not currently support for encrypted channels'
```
kn100: support(*ed)

WillSewell: Another good spot! Fixed

PR #293, opened 2018-08-03
Rename `encrypted` to `forceTLS`

WillSewell:
## What does this PR do?

Rename `encrypted` to `forceTLS`. This keeps the `encrypted` around, but it is deprecated.

We are doing this because `encrypted` is easily confused with our end-to-end encryption feature.

## Checklist

- [x] All tests are passing.


src/runtimes/web/default_strategy.ts
```diff
@@ -17,16 +17,16 @@ var getDefaultStrategy = function(config : any) : any {
...
       httpPath: config.wsPath
     }],
     [":def", "wss_options", [":extend", ":ws_options", {
-      encrypted: true
+      useTLS: true
     }]],
     [":def", "sockjs_options", {
-      hostUnencrypted: config.httpHost + ":" + config.httpPort,
-      hostEncrypted: config.httpHost + ":" + config.httpsPort,
+      hostNonTLS: config.httpHost + ":" + config.httpPort,
```
kn100: From what I can tell this is a breaking API change for customers who were using the hostEncrypted/hostUnencrypted options - was this intentional?

WillSewell: This is ok actually, the public config options are https://github.com/pusher/pusher-js#wshost-wsport-wssport-httphost-httpport-httpsport, and don't need to change.

`hostNonTLS` and `hostTLS` are purely internal.

WillSewell:
@kn100 should be in a good state now. Let me know if it's ok to merge.

PR #323, opened 2019-01-29
Support receipt of user_id in on client_events on presence channels

leesio:
## What does this PR do?

If a `user_id` is present in the top level of a client event on a presence channel, emit it as a `user_id` key in a metadata object. `Channels` always call callbacks with a `metadata` object as the 2nd argument

While implementing this feature I noticed some peculiarities with the types/message semantics so tried to tidy that up too: We receive a [MessageEvent](https://developer.mozilla.org/en-US/docs/Web/API/MessageEvent) on the websocket and decode that into a [PusherEvent](https://pusher.com/docs/pusher_protocol#events)

## Checklist

- [x] All tests are passing.


README.md
```diff
@@ -417,6 +417,14 @@ channel.bind('my-event', function () {
 }, { name: 'Pusher' });
 ```
 
+For client-events on presence channels, bound callbacks will be called with an additional argument, an object containing the `user_id` of the user who triggered the event
```
hosamaly: I believe that camel-case naming is more conventional in the JavaScript world, so I suggest renaming this to `userId`.

README.md
```diff
@@ -417,6 +417,14 @@ channel.bind('my-event', function () {
...
+
+```
+presenceChannel.bind('client-message', function (data, metadata) {
```
hosamaly:
```suggestion
presenceChannel.bind('event-name', function (data, metadata) {
```

spec/javascripts/unit/core/channels/presence_channel_spec.js
```diff
@@ -214,10 +235,25 @@ describe("PresenceChannel", function() {
...
+          var callback = jasmine.createSpy("callback");
+          channel.bind("client-something", callback);
+
+          channel.handleMessage({
+            event: "client-something",
+            data: 9,
+            user_id: "abc-def",
+          });
+
+          expect(callback).toHaveBeenCalledWith(9, {user_id: "abc-def"});
```
hosamaly: Doesn't the callback know about the event name?

src/core/channels/presence_channel.ts
```diff
@@ -41,12 +42,28 @@ export default class PresenceChannel extends PrivateChannel {
...
+    if (event.indexOf("pusher_internal:") === 0) {
+      this.handleInternalMessage(message)
+    } else {
+      var data = message.data;
+      if (message.user_id) {
+        var metadata = {};
+        metadata['user_id'] = message.user_id;
+        this.emit(event, data, metadata);
+      } else {
+        this.emit(event, data);
```
hosamaly: I think that it's preferable to pass in an empty hash for the metadata in all cases, avoiding any possible errors from `undefined`.

src/core/events/dispatcher.ts
```diff
@@ -62,7 +62,11 @@ export default class Dispatcher {
     var callbacks = this.callbacks.get(eventName);
     if (callbacks && callbacks.length > 0) {
       for (i = 0; i < callbacks.length; i++) {
-        callbacks[i].fn.call(callbacks[i].context || global, data);
+        if(metadata) {
```
hosamaly:
```suggestion
        if (metadata) {
```

src/core/events/dispatcher.ts
```diff
@@ -62,7 +62,11 @@ export default class Dispatcher {
...
+          callbacks[i].fn.call(callbacks[i].context || global, data, metadata);
+        } else {
+          callbacks[i].fn.call(callbacks[i].context || global, data);
```
hosamaly: As far as I understand, this is the same as the expression on line 66, so I suggest removing the `if` condition and the `else` block.

README.md
```diff
@@ -417,6 +417,14 @@ channel.bind('my-event', function () {
...
+  console.log('received message', data.message, 'from', metadata.user_id);
```
hosamaly: What does `data` look like? I couldn't figure out whether it would always have a `message` field...

src/core/channels/presence_channel.ts
```diff
@@ -41,12 +42,28 @@ export default class PresenceChannel extends PrivateChannel {
...
+      }
+    }
+  }
+  handleInternalMessage(message: Message) {
+    var event = message.event;
+    var data = message.data;
     switch (event) {
       case "pusher_internal:subscription_succeeded":
```
hosamaly: This code is a duplicate of `Channel.handleMessage`, except for the call to `members.onSubscription`. Can we reduce the duplication?

src/core/channels/channel.ts
```diff
@@ -55,12 +56,13 @@ export default class Channel extends EventsDispatcher {
...
+  /** Handles a message. For internal use only.
    *
-   * @param {String} event
-   * @param {*} data
+   * @param {Message} message
    */
-  handleEvent(event : string, data : any) {
+  handleMessage(message: Message) {
+    var event = message.event;
+    var data = message.data;
```
hosamaly: Can we use `const` (or `let`) instead of `var`?

README.md
petarslavnic: Data comes from a triggered event which name, in this case, is `client-message` and could be the type of object with a defined key-value pair.

README.md
petarslavnic: There are missing Code of Conduct. Some lint rules would be nice. Until then ¯\_(ツ)_/¯

README.md
petarslavnic: Client events must be prefixed by client-. Events with any other prefix will be rejected by the Channels server, as will events sent to channels to which the client is not subscribed.

https://pusher.com/docs/client_api_guide/client_events#trigger-events

spec/javascripts/unit/core/channels/presence_channel_spec.js
petarslavnic: Nope. The callback has been executed by that event.

petarslavnic:
@leesio why do you need `metadata` when you can pass `user_id` trough `data` object?

leesio:
@petarslavnic there are a couple of reasons:

1. The `data` object isn't always an `Object`. It can be a `String` or an `Array. We could force it to be an object, but that would be a breaking changes for existing users.
2. The `data` argument (whatever it is) is sent by the pusher-server library consumer. I don't think we should be altering that at all.

petarslavnic:
@leesio The data argument is a value that we send when triggering event on a presence channel. We expect what we sent and that is an Object in our case.

petarslavnic:
@leesio My point is that we don't need that info 
as metadata because we can pass it through data value if we want to.

> @leesio The data argument is a value that we send when triggering event on a presence channel. We expect what we sent and that is an Object in our case.

My point is that we don't need that info 
as metadata because we can pass it through data value if we want to.

leesio:
@petarslavnic this has some advantages over what you're describing.

The client _could_ send the `user_id` parameter as part of the message, but that leaves it to the client to be 'honest' about who they are.

The `user_id` provided here is extracted from the token used to authenticate to the presence channel. This token is hashed by the server library, so it shouldn't be possible for the client to modify it.

Ofcourse, if you'd rather send the `user_id` with the data, that's fine, you can just ignore the `metadata` argument

README.md
leesio: It wouldn't always have a message field. It could be an `Object` an `Array` or a `String`. I should update this example as it could be mislreading

README.md
leesio: @hosamaly I agree that the camelCase would be more idiomatic and that the inconsistency here is not good.

In this specific case, I'm trying to be consistent with other presence events though. For example:

```
Pusher : Event recd : {"event":"pusher_internal:member_added","data":{"user_id":"user-2019-02-01T09:58:09.388Z"}
```

src/core/channels/presence_channel.ts
leesio: Agreed

src/core/channels/presence_channel.ts
leesio: Tried to reduce the duplication with https://github.com/pusher/pusher-js/pull/323/commits/86936530cbb2d582c16e51c5b854ecedd0d7ca7c

src/core/channels/channel.ts
leesio: We could and ultimately should, but I think this is a broader question about adopting modern JS conventions and a more modern transpiling step. Not something we should necessarily start doing now.

src/core/events/dispatcher.ts
leesio: I changed this _slightly_ https://github.com/pusher/pusher-js/pull/323/commits/9ef0f7b90018a6adb36f04373748ee6fe326f75a.

I think using an `args` array is nicer and doesn't break any tests!

src/core/channels/channel.ts
```diff
@@ -55,35 +56,42 @@ export default class Channel extends EventsDispatcher {
...
-        this.subscriptionPending = false;
-        this.subscribed = true;
-        if (this.subscriptionCancelled) {
-          this.pusher.unsubscribe(this.name);
-        } else {
-          this.emit("pusher:subscription_succeeded", data);
-        }
-      }
-    } else {
+  handleMessage(message: Message) {
```
WillSewell: I'm not sure about the change of name and signature:

* I think "Event" is actually a better name than "Message" because it's what they are referred to in our docs: https://pusher.com/docs/pusher_protocol#events. I think it's the `Message` type that is poorly named.
* `Message` defines the wire format of Pusher events, and I'm concerned about it bubbling so far up the call stack. I would prefer these `handleMessage` methods to be called with only the parameters they need, e.g. `handleMessage` for public channels does not need to be called with a `user_id`. It also feels wrong that non pusher protocol events are being wrapped in a `Message`, e.g. https://github.com/pusher/pusher-js/pull/323/files#diff-0e82fa5891d842a9141e2aad30757dd7R90

src/core/channels/encrypted_channel.ts
```diff
@@ -46,14 +47,15 @@ export default class EncryptedChannel extends PrivateChannel {
...
-  handleEvent(event: string, data: any) {
+  handleMessage(message: Message) {
+    var event = message.event;
+    var data = message.data;
     if (event.indexOf("pusher_internal:") === 0 || event.indexOf("pusher:") === 0) {
-      super.handleEvent(event, data);
+      super.handleMessage(message);
       return
     }
     this.handleEncryptedEvent(event, data)
```
WillSewell: If `handleEvent` is being renamed and the signature changed, then I think this should be too.

src/core/events/dispatcher.ts
```diff
@@ -52,17 +52,21 @@ export default class Dispatcher {
...
-  emit(eventName : string, data?: any) : Dispatcher {
+  emit(eventName : string, data?: any, metadata?: any) : Dispatcher {
     var i;
 
     for (i = 0; i < this.global_callbacks.length; i++) {
       this.global_callbacks[i](eventName, data);
     }
 
     var callbacks = this.callbacks.get(eventName);
+    var args = [data];
```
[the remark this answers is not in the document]
leesio: Assume you mean `data`?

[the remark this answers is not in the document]
WillSewell: Ah, yes I did.

src/core/channels/presence_channel.ts
```diff
@@ -52,10 +52,17 @@ export default class PresenceChannel extends PrivateChannel {
...
-      if (message.user_id) {
+
+      if (event.indexOf("client-") === 0) {
+
+        // client events on presence channels include a user_id field in the top
+        // level of the message object. This user_id parameter should be
+        // included in a metadata object included as the second argument to
+        // bound callbacks
         var metadata = {};
         metadata['user_id'] = message.user_id;
```
hosamaly: If we make this configurable via a server-side switch then this field could be `null`.

src/core/channels/presence_channel.ts
```diff
@@ -66,15 +73,14 @@ export default class PresenceChannel extends PrivateChannel {
...
-        } else {
+        if (!this.subscriptionCancelled) {
           this.members.onSubscription(data);
-          this.emit("pusher:subscription_succeeded", this.members);
         }
-        break;
+        this.handleSubscriptionSucceededMessage({
+          event: event,
+          data: this.members,
+        });
```
hosamaly: I think that we should do this _before_ calling `members.onSubscription`. That way, we make sure that the internal state is consistent and we know how to proceed if a member does something unexpected (e.g. unsubscribes).

src/core/channels/presence_channel.ts
```diff
@@ -84,9 +90,7 @@ export default class PresenceChannel extends PrivateChannel {
         if (removedMember) {
           this.emit('pusher:member_removed', removedMember);
         }
-        break
-      default:
```
hosamaly: Shouldn't we have  a `default` case, at least to log that something unexpected was received?

src/core/channels/presence_channel.ts
leesio: I think we need to update the state of `members` before emitting an `event` containing the members!

src/core/channels/presence_channel.ts
leesio: Elsewhere, we don't log/expose anything to do with unhandled `pusher_internal` events. I can see a case for calling `super` in case any `internal` events are handled at the `Channel` level, but in this case I think it's consistent.

src/core/channels/presence_channel.ts
hosamaly: You're right. My comment was actually misleading. Let me try again.

I think that we should update `subscribed` and `subscriptionPending` before we call `members.onSubscription`.

Does this make sense?

leesio: We could certainly do that. I initially moved this around so that we hanlded subscription state in only one place (the `Channel` class). Partly inspired by your comment about duplication

> This code is a duplicate of `Channel.handleMessage`, except for the call to `members.onSubscription`. Can we reduce the duplication?

leesio: @hosamaly 1f9fbc70a8306da5d3a0bfa8d368061ba7cc516c this is the alternative imo. I don't think it's terrible to update the membership before updating the internal state, but I suppose this is more 'correct'

src/core/channels/presence_channel.ts
leesio: This is no longer relevant 👍

PR #325, opened 2019-02-14
R4.4

leesio:
## What does this PR do?

[Description here]

## Checklist

- [ ] All tests are passing.


README.md
```diff
@@ -417,6 +417,14 @@ channel.bind('my-event', function () {
 }, { name: 'Pusher' });
 ```
 
+For client-events on presence channels, bound callbacks will be called with an additional argument. This argument is an object containing the `user_id` of the user who triggered the event
+
+```
```
WillSewell: Suggest suffixing this with `js`

leesio: Not sure I follow?

PR #338, opened 2019-03-22
Replace sockjs module with the root repo rather than a fork

WillSewell:
## What does this PR do?

git://github.com/pusher/sockjs-client.git was removed, and the
changed have been moved into git://github.com/sockjs/sockjs-client.git,
see: https://github.com/sockjs/sockjs-client/commit/5b91dbc0e217a9cd959f915fec090370b29878da

## Checklist

- [x] All tests are passing.


WillSewell:
I don't think we want to do this as that commit isn't referenced by a branch in git://github.com/pusher/sockjs-client.git.

Issue #298, opened 2018-08-15
Tree shaking unused features


marksamman:
Would it be possible to organize this package so that the E2E-encryption implementation isn't bundled by tree-shaking capable bundlers if it's not used? My projects are not using end-to-end encrypted channels and upgrading from 4.2.2 to 4.3.0 resulted in my JS bundles increasing by 60KB post-minification (UglifyJS). It is significant enough that I will pin my projects to 4.2.2 for now.

leesio:
Hi @marksamman, thanks for bringing this up. This is something we're considering at the moment.

I think we'd have to update the build process to enable tree shaking, but we're not happy with the increase int he bundle size either.

joker-777:
That would be great! I think the pusher package

![screenshot_20190118_122754](https://user-images.githubusercontent.com/378859/51402814-912a1800-1b1c-11e9-83ad-ca0445c8b165.jpeg)
 is too big.

stof:
As the E2E encryption is based on the prefix of the channel name, I don't see any way to tree-shake it though

stof:
Idea to allow tree-shaking in a new major version of the library (as that involves BC breaks): expose an API like `pusher.subscribe(new Pusher.Channel(name))`, exposing `Channel`, `PrivateChannel`, `PresenceChannel` and `EncryptedChannel` as exports.
Then, if a module build is provided, this would allow tree-shaking to happen (webpack is much better at tree-shaking ESM builds than CJS ones).

leesio:
As @stof says, it's unlikely we can solve the bundle size increase with tree shaking. We're planning on splitting the end to end encryption functionality into a separate helper library at some stage.

I created an issue to track that #403

In the meantime, I'm going to close this issue.

PR #340, opened 2019-04-04
Refactor the creation of the default strategy

### LINKED ISSUE #244: Initializing Pusher blocks the UI for more than 100ms
I'm using the latest version of Pusher (4.1 loaded from your CDN).

when I instantiate pusher in my code (`new Pusher(API_KEY, {encrypted: true})`), this blocks the UI for more than 100ms (I'm running profiling with a 5x CPU throttling enabled in the Chromium DevTools, to simulate a computer less powerful than mine). During this time, the UI is unresponsive (and none of my code can reduce this time except dropping the usage of Pusher entirely, which is not good).
A bit more than half of the time is spent in `createConnectionManager`, most of it being spent by `ConnectionManager#updateStrategy`.

It would be great if this code could be optimized (or broken down into several blocks, to allow the UI to respond sooner)

stof:
## What does this PR do?

Instead of using an interpreted DSL which cannot be optimized by JS engines, the JS calls are done directly in place. This is a lot faster. Closes #244 
I did some manual measurements of the time needed to run `new Pusher()`, and both Chrome and Firefox show a 10x improvement on my desktop (going from ~30ms to ~3ms, which will make it less likely to exhaust the frame budget of 16ms).

## Checklist

- [x] All new functionality has tests. (no new feature)
- [x] All tests are passing.
- [x] New or changed API methods have been documented. (no new public API)

## Future improvements

- Improving type safety by typing the `config` argument properly, and specifying the actual signature of the `defineTransport` callback (but there is already much more type safety than before, which is also how I discovered that the CachedStrategy actually had broken types).
- decide what the proper way to handle this `defineTransport` is. My initial implementation had it inlined inside `defineTransportStrategy`, at the expense of duplicating the logic in isomorphic and web. But then, I needed it for the `transport_lists` spec, as it is covering this logic. This is why I moved it in core. And I pass it as a callback argument rather than importing it to avoid circular references (this `defineTransport` needs access to `Runtime.Transports` but importing the runtime inside the module defining the default strategy of the runtime will break). I haven't found a nicer way.
- potentially simplify the `SequentialStrategy` as it is always used with a single sub-strategy
- potentially simplify the `IfStrategy` as it is always use with a `isSupported()` check as test callback (and so it could become a `SupportingStrategy` accepting a tested strategy as first argument instead)


leesio:
@stof this is really interesting. Thanks a lot for the testing, and the implementation, I'm going to have a look shortly.

Out of interest, were you having particular issues with the performance of pusher-js?

stof:
Well, my issue is described in https://github.com/pusher/pusher-js/issues/244

I was trying to optimize my app, and spending a long time blocking the main thread due to `new Pusher()` was not good. In my latest measurement (on a powerful desktop not using any CPU throttling), initializing the whole JS of my app tool 225ms. And then, in the next run of the event loop, setting up Pusher took 40ms (I stopped doing that directly in the initialization years ago when I reported the other, to give a change for the page to be rendered a bit earlier). Spending that much time initializing Pusher is not good for performance.

The other room for improvement I see is the size of the library, but this is discussed in #298

leesio:
@stof the size of the library is definitely something that concerns me too. We're discussing making the end-to-end encryption functionality 'opt-in' and providing bundles with and without.

For the moment, are you using your fork of pusher-js in your application?

stof:
not yet. I only tried it locally.

leesio:
In general, I'm in favour of your approach over the existing implementation.

Given what you say about the speed of instantiation it's more performant and I think it's easier to understand from a newcomers POV.

I'll need to review the PR in more detail, but in principle I'm happy with this approach!

stof:
@leesio any news about this review ?

leesio:
Hey @stof,
I reviewed this PR this morning. Nice work!
That part of the library is (in my opinion) needlessly complex so I'm impressed you managed to make sense of it at all, let alone achieve the kind of performance improvements you describe 👍 

I think your implementation is definitely the way I'd like to take the strategy building. There are some changes I'd like to make to the implementation though, mainly for the purposes of reducing the complexity of the code and improving readability and maintanability. This is by no means a criticism of your work, I realise this complexity existed before your changes!

At the moment, I'm in favour of including your changes in the next release (as is) though I'd like to discuss with a couple of colleagues first.

I hope this delay doesn't cause you problems.

stof:
Well, for my work, I explicitly decided to *only* convert the existing DSL to the equivalent JS code, without refactoring it more, to keep the PR focused on that (as I figured out that the DSL itself was the performance issue).
I totally agree with you that more work could be done here to reduce complexity, but I think it can be a next step (so I totally agree with you on merging that refactoring as is and doing the simplification separately).

The delay is not causing me problems (my app runs in prod since years with this DSL-based instantiation, so I don't have a big blocker to deploy the faster version with a hard deadline). But I want to make sure this does not get forgotten, given the performance impact it has.

stof:
@leesio any news on the stats of this PR ?

leesio:
Hey @stof 
At the moment, I'm starting to merge PR's updating the build process deps and making the tests easier to run. After those housekeeping changes this PR is on the top of my list for merge/release.

Thanks for your patience with this one, hopefully the changes to the build/testing should make similar changes quicker to merge in the future

leesio:
Superseded by #360

PR #348, opened 2019-05-28
Netinfo test build

leesio:
## What does this PR do?

[Description here]

## Checklist

- [ ] All tests are passing.


src/runtimes/react-native/net_info.ts
```diff
@@ -1,4 +1,4 @@
-import {NetInfo as NativeNetInfo} from 'react-native';
+import {default as NativeNetInfo} from '@react-native-community/netinfo';
```
oblador:
```suggestion
import NativeNetInfo from '@react-native-community/netinfo';
```

package.json
```diff
@@ -28,7 +28,7 @@
   },
   "homepage": "https://github.com/pusher/pusher-js",
   "devDependencies": {
-    "es3ify-loader": "^0.2.0",
+    "@react-native-community/netinfo": "^3.2.0",
```
oblador: `peerDependency`?

leesio: I actually think this should be a `devDependency` *and* a `peerDependency`, though I'm interested in your thoughts too, @oblador.

The js part of the `netinfo` module is included in the built artefact (`dist/react-native/pusher.js`) so it is required for the react-native build so is a `devDependency`

The native parts of the module need to be installed and linked in the main project so it _should_ be a `peerDependency` too. The downside of declaring it as a `peerDependency` is that node/web users get a warning about a missing dependency that they don't actually need.

I'm tempted to keep it as a `devDependency` and update the documentation to warn users that there's an 'unofficial' `peerDependency`.

Do you have a preference, @oblador ?

src/runtimes/react-native/net_info.ts
oblador: It's typescript, never mind :-)

package.json
oblador: I didn't understand that you vendored the netinfo module. Sounds kinda dangerous to me as the JS and native code are tightly coupled. This would make upgrading netinfo difficult for the end user, especially if there's no `peerDependency` to warn about incompatibilities. 

I would favor breaking this module up into a RN specific module instead.

leesio: I agree that bundling the js component of `NetInfo` is unpleasant.

Looking at the old build process, we leverage the `externals` option in webpack to avoid bundling any of the react-native module code in the bundle. I think this may be a good approach here too.

It tells webpack not to include those modules in the bundle, but assume they're available at runtime. This way both the js and the native parts of `netinfo` come from the application's dependency.

I added an example build using this approach in f07bba5a5312731fa33e52bd0863b1fd012f64dc

if you can confirm this still works in your application. I can push on getting this approach released

brandon-dewitt:
@oblador as of RN 0.60.0 NetInfo is removed.  Are you ok with the changes @leesio made to exclude NetInfo from the bundle?  I would love to help get this complete if there is anymore work that needs to be done.

oblador:
@brandon-dewitt Yes I am, was on vacation earlier so sorry for the lack of responses.

brandon-dewitt:
Awesome, thanks for the update @oblador !

leesio:
I'm closing this PR. It was always temporary and the proposed change has been included in the latest realease.

PR #360, opened 2019-07-12
Refactor the creation of the default strategy (rebased)

### LINKED ISSUE #244: Initializing Pusher blocks the UI for more than 100ms

leesio:
*I created this PR to replace #340.  I rebased the branch from the latest master. The below is the original pr description written by @stof*

## What does this PR do?

Instead of using an interpreted DSL which cannot be optimized by JS engines, the JS calls are done directly in place. This is a lot faster. Closes #244 
I did some manual measurements of the time needed to run `new Pusher()`, and both Chrome and Firefox show a 10x improvement on my desktop (going from ~30ms to ~3ms, which will make it less likely to exhaust the frame budget of 16ms).

## Checklist

- [x] All new functionality has tests. (no new feature)
- [x] All tests are passing.
- [x] New or changed API methods have been documented. (no new public API)

## Future improvements

- Improving type safety by typing the `config` argument properly, and specifying the actual signature of the `defineTransport` callback (but there is already much more type safety than before, which is also how I discovered that the CachedStrategy actually had broken types).
- decide what the proper way to handle this `defineTransport` is. My initial implementation had it inlined inside `defineTransportStrategy`, at the expense of duplicating the logic in isomorphic and web. But then, I needed it for the `transport_lists` spec, as it is covering this logic. This is why I moved it in core. And I pass it as a callback argument rather than importing it to avoid circular references (this `defineTransport` needs access to `Runtime.Transports` but importing the runtime inside the module defining the default strategy of the runtime will break). I haven't found a nicer way.
- potentially simplify the `SequentialStrategy` as it is always used with a single sub-strategy
- potentially simplify the `IfStrategy` as it is always use with a `isSupported()` check as test callback (and so it could become a `SupportingStrategy` accepting a tested strategy as first argument instead)


stof:
the PR number is wrong in the first sentence. #360 is this PR. The replaced PR is #340

leesio:
@stof indeed. My mistake. Updated

leesio:
@stof I think I'm in a good position to merge this PR now. I'll be including it in the next release.

I really appreciate the work you put into it aswell as your patience in waiting for it to be merged/released

stof:
great :champagne:

stof:
@leesio do you have an approximate date for the next release ?

leesio:
@stof should be very soon. I wanted to release it today, but ran out of time.

leesio:
@stof I just pushed this new release this morning.

PR #370, opened 2019-07-24
Optimize build

stof:
## What does this PR do?

This optimize the build to reduce the bundle size:

- it uses ES module as the output of Typescript (and so the input of webpack optimizations), as some of the most important optimizations performed by webpack are applied only on ES modules (see https://webpack.js.org/plugins/module-concatenation-plugin/)
- ~~it removes the useless Buffer polyfill injected by webpack due to tweetnacl using it (as it is using it only in node.js environments and has a different code path for the browser)~~ now extracted to #375

Here are the size of files for the web bundles (using `ls -shS1 dist/web/`, and keeping only the pusher files as sockjs and json2 are irrelevant for this PR):

before:
```
272K pusher.js
128K pusher.min.js
```

after first optimization (TS modules):

```
260K pusher.js
124K pusher.min.js
```

after both optimizations:
```
208K pusher.js
100K pusher.min.js
```

### DIFF
--- a/spec/javascripts/helpers/mocks.js
+++ b/spec/javascripts/helpers/mocks.js
@@ -64,14 +64,6 @@ var Mocks = {
     return request;
   },
 
-  getDependencies: function() {
-    return {
-      load: jasmine.createSpy("load"),
-      getRoot: jasmine.createSpy("getRoot"),
-      getPath: jasmine.createSpy("getPath")
-    };
-  },
-
   getJSONPSender: function() {
     return {
       send: jasmine.createSpy("send")
--- a/spec/javascripts/helpers/pusher_integration.js
+++ b/spec/javascripts/helpers/pusher_integration.js
@@ -1,0 +1,1 @@
+module.exports = require('./pusher_integration_class').default;
--- a/spec/javascripts/helpers/pusher_integration_class.ts
+++ b/spec/javascripts/helpers/pusher_integration_class.ts
@@ -1,7 +1,7 @@
 import Pusher from '../../../src/core/pusher';
 import {ScriptReceiverFactory} from '../../../src/runtimes/web/dom/script_receiver_factory';
 
-class PusherIntegration extends Pusher {
+export default class PusherIntegration extends Pusher {
 
   static Integration : any = {
     ScriptReceivers: new ScriptReceiverFactory(
@@ -10,5 +10,3 @@ class PusherIntegration extends Pusher {
     )}
 
 }
-
-module.exports = PusherIntegration;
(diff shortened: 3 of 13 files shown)


stof:
I don't know what is failing on Travis. I don't get any failure message except `it failed`

leesio:
Hey @stof. Thanks for the PR.

I'm fine with both of these changes in principle. Removing the `node` buffer polyfill seems like a quick win. Nice spot!

It looks like the build is failing 'cos some of the unit tests are broken by changing the `module` to `es6`. It's not immediately clear to me why. I'll try to look into this soon, but in the meantime, how would you feel about just including just the change to remove the `Buffer` polyfill?

stof:
I think the issue is related to the test trying to mock `DependenciesModule.Dependencies`. Mocking a property of a commonjs export works fine. But it might not behave the same when trying to mock an ES6 export by reassigning the property (because webpack probably reference the export directly in the code elsewhere).

stof:
@leesio I opened #375 with only the Buffer polyfill removal

leesio:
> I think the issue is related to the test trying to mock DependenciesModule.Dependencies. Mocking a property of a commonjs export works fine. But it might not behave the same when trying to mock an ES6 export by reassigning the property (because webpack probably reference the export directly in the code elsewhere).

Yeah, it does look like that. I didn't have chance to look into it in depth.

Thanks for creating the other PR. I'll try to get that merged and deployed today.

stof:
@leesio I found a way to refactor some of the tests to make them pass with ES modules. But I still have issues with connection_spec.js and handshake_spec.js, which both try to mock functions exported by the `core/connection/protocol/protocol` module (and I cannot use the same trick here, as functions are the export themselves, not methods on an exported object).

stof:
A solution here could be to export an object as default export for Protocol instead, as usages always look like `import * as Protocol from '../protocol/protocol'` (so named exports are never used directly, but always as methods). What do you think ?

stof:
@leesio ?

leesio:
Hey @stof, apologies I still haven't had a lot of time to look into this issue. I should have some time this week.
At first glance I'm not against your suggested approach!

stof:
@leesio I applied my change, and the build is now green locally.

To review the second commit refactoring the `protocol` module, I suggest you to ignore space changes, as most modified lines are only about increasing the indentation.

stof:
I fixed the issue with integration tests. And as #375 is now merged, here are the updated stats for this optimization.

Before

```
216K pusher.js
108K pusher.min.js
```

After:

```
208K pusher.js
100K pusher.min.js
```

leesio:
Superceded by #383

PR #375, opened 2019-08-19
Remove the useless Buffer polyfill from the build

stof:
## What does this PR do?

Optimize the bundle size.
tweetnacl works without Buffer when atob is available (which is the case for web runtimes). So there is no reason to include the polyfill.

This is extracted from #370

### DIFF
--- a/webpack/config.shared.js
+++ b/webpack/config.shared.js
@@ -26,6 +26,12 @@ module.exports = {
       { test: /\.ts$/, loader: 'ts-loader' },
     ]
   },
+  node: {
+    // nacl uses Buffer on node.js but has a different code path for the browser.
+    // We don't need webpack to include a Buffer polyfill when seeing the usage,
+    // as it won't be used.
+    Buffer: false
+  },
   plugins: [
     new webpack.BannerPlugin({banner:  banner, raw: true}),
     new webpack.DefinePlugin({


stof:
the build for master is red too.

leesio:
@stof I pushed your commit to a new branch and created a new PR and the build seems to work. The only thing I can think right now is that PRs from forks can't access the browserstack creds that travis needs to run the safari tests

stof:
@leesio if this is using a secure env variable, this is indeed the case. Maybe the browserstack-based jobs should be skipped in case creds are not available.

leesio:
@stof yeah that shouldn't be too much effort. We're only using browserstack for safari at the mo. The tests would still run in Headless Chrome/Firefox which is still a good indicator of correctness.

> the build for master is red too. It passes locally
As a general point, I have spent some time trying to make the tests more deterministic and I think they're better now, but unfortunately they're still not perfect. It's not unusual for test runs to fail and then succeed after a retry :(

leesio:
I'm closing this since I merged #376

leesio:
@stof I'm afraid I won't get a chance to push a new release today. I'm on holiday next week but I'll push a new release as soon as I'm back in the office.
Sorry for the delay!

leesio:
@stof just FYI: I published a new release yesterday without the buffer polyfill

stof:
great thanks.

stof:
Note that projects using webpack as their bundler will need to do the same disabling. Otherwise, their own webpack bundling would add a Buffer polyfill (for the same reason).

Things will be great in webpack 5 when this node polyfilling will be opt-in rather than opt-out (based on the fact that most packages aimed at being used for frontend are already taking care of compat properly rather than expecting node APIs to be available).

Issue #382, opened 2019-09-05
Warning: RCNNetInfo - You are using the deprecated API


Deg5112:
pusher-js lib appears to be calling a deprecated method from react-native-netinfo.

`Warning: RCNNetInfo - You are using the deprecated API...`

At a base level test case
literally just having the latest installed: "@react-native-community/netinfo": "^4.2.0"
and having a file with nothing in it other than import Pusher from 'pusher-js/react-native': "5.0.1"

i will get the warning, when deploying to ios/android simulator

i dug a little into the "@react-native-community/netinfo" specifically https://github.com/react-native-community/react-native-netinfo/blob/master/src/index.ts
line 144:
```
/**
 * Deprecated method to get the current state. You should upgrade to the new `fetch` method and
 * handle the new state type.
 *
 * @deprecated
 */
export function getConnectionInfo(): Promise<DeprecatedTypes.NetInfoData> {
  DeprecatedUtils.warnOnce();
  return DeprecatedState.latest();
}
```

Will the next update call the new fetch method, or am i missing something? Thanks!

Deg5112:
<img width="1026" alt="React_Native_Debugger_-_Connected__port_19001_" src="https://user-images.githubusercontent.com/13839383/64406445-abc2ea00-d036-11e9-96fb-9a188add7760.png">

leesio:
@Deg5112 thanks for letting us know about this. I understood when `NetInfo` was removed from `react-native` core that the old API would be supported for a while. Apparently I underestimated how long a while is in the `react-native` world.

I had a quick look and it seems like the new API works OK. I wonder if you'd mind trying a custom build to see if it solves your problem? You can install the custom build with

```
npm install git+ssh://git@github.com/pusher/pusher-js.git#use-new-netinfo-api
```
If this is all it takes to resolve this, it should be fine to include it in the next release

Deg5112:
Thanks for the reply james. I installed the package at the branch you specified with that url. However, it looks like it's still calling the same deprecated getConnectionInfo method, and i still get the warning. Full stack trace below

Warning: RNCNetInfo - You are using the deprecated API. It will still work, but you must upgrade to the new API to receive the new features. The old API will be removed in the future
- node_modules/expo/build/environment/muteWarnings.fx.js:17:23 in warn
- node_modules/@react-native-community/netinfo/src/internal/deprecatedUtils.ts:79:4 in warnOnce
[...]

PR #384, opened 2019-09-23
Update net_info.ts

subotkevic:
Source: https://github.com/react-native-community/react-native-netinfo/issues/123

This should get rid of Deprecation warning (https://github.com/pusher/pusher-js/issues/382)


leesio:
Ahh thanks for that @subotkevic, I missed this fairly subtle API change. I'll merge this into my test branch and see about merging that PR

PR #399, opened 2019-11-07
Error logging improvements

WillSewell:
## What does this PR do?

There were three objectives of this PR:

1. If the auth response body is not JSON and logToConsole is set, it should log an error (not just mention no callback is bound)
2. All calls to Logger.warn logs should use console.warn if no default logger is set
3. Extend Logger to include an error method that calls console.error. Error logs should be updated to use this.

I made a couple of other changes while I was at it. The commit messages should be self-explanatory.

There are a couple of points I'd like feedback on in particular:

1. Connection errors are logged at warning level. This is because expected conditions like disconnects are included here. However it also includes error like failure to decode websocket messages, which I think should ideally be logged as errors.
2. @leesio it would be good to get feedback on how I refactored the `Logger.ts` file. I wasn't to implement something like private methods, but I'm not sure this best way of doing this.

## Checklist

- [x] All tests are passing.

### DIFF
--- a/src/core/channels/channel.ts
+++ b/src/core/channels/channel.ts
@@ -96,7 +96,8 @@ export default class Channel extends EventsDispatcher {
     this.subscriptionCancelled = false;
     this.authorize(this.pusher.connection.socket_id, (error, data)=> {
       if (error) {
-        this.emit('pusher:subscription_error', data)
+        Logger.warn('Error', data);
+        this.emit('pusher:subscription_error', data);
       } else {
         this.pusher.send_event('pusher:subscribe', {
           auth: data.auth,
--- a/src/core/channels/channel.ts
+++ b/src/core/channels/channel.ts
@@ -96,7 +96,7 @@ export default class Channel extends EventsDispatcher {
     this.subscriptionCancelled = false;
     this.authorize(this.pusher.connection.socket_id, (error, data)=> {
       if (error) {
-        Logger.error('Error', data);
+        Logger.error(data);
         this.emit('pusher:subscription_error', data);
       } else {
         this.pusher.send_event('pusher:subscribe', {
--- a/src/core/channels/channel.ts
+++ b/src/core/channels/channel.ts
@@ -96,6 +96,9 @@ export default class Channel extends EventsDispatcher {
     this.subscriptionCancelled = false;
     this.authorize(this.pusher.connection.socket_id, (error, data)=> {
       if (error) {
+        // Why not bind to 'pusher:subscription_error' a level up, and log there?
+        // Binding to this event would cause the warning about no callbacks being
+        // bound (see constructor) to be suppressed, that's not what we want.
         Logger.error(data);
         this.emit('pusher:subscription_error', data);
       } else {
--- a/src/core/channels/channel.ts
+++ b/src/core/channels/channel.ts
@@ -96,7 +96,7 @@ export default class Channel extends EventsDispatcher {
     this.subscriptionCancelled = false;
     this.authorize(this.pusher.connection.socket_id, (error, data)=> {
       if (error) {
-        Logger.warn('Error', data);
+        Logger.error('Error', data);
         this.emit('pusher:subscription_error', data);
       } else {
         this.pusher.send_event('pusher:subscribe', {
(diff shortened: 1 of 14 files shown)


src/core/channels/encrypted_channel.ts
```diff
@@ -32,7 +32,7 @@ export default class EncryptedChannel extends PrivateChannel {
       if (!sharedSecret) {
         let errorMsg = `No shared_secret key in auth payload for encrypted channel: ${this.name}`;
         callback(true, errorMsg);
-        Logger.warn(`Error: ${errorMsg}`);
+        Logger.error(errorMsg);
```
leesio: Is this necessary now? We log the error in the callback anyway?

src/core/logger.ts
```diff
@@ -1,24 +1,46 @@
...
+
+function globalLogError(message) {
+  if (global.console && global.console.error) {
+    global.console.error(message);
+  } else {
+    globalLogWarn(message);
+  }
+}
+
+function log(defaultLoggingFunction, ...args: any[]) {
```
leesio: Not sure I understand the name `defaultLoggingFunction`. This argument also doesn't have a type signature.

src/core/logger.ts
```diff
@@ -1,24 +1,46 @@
 import { stringify } from './utils/collections';
 import Pusher from './pusher';
 
+function globalLog(message) {
```
leesio: Is `message` a `string`? There's no type signature

leesio:
Also any reason not to run the formatter?

src/core/channels/channel.ts
```diff
@@ -96,7 +96,8 @@ export default class Channel extends EventsDispatcher {
     this.subscriptionCancelled = false;
     this.authorize(this.pusher.connection.socket_id, (error, data)=> {
       if (error) {
-        this.emit('pusher:subscription_error', data)
+        Logger.error(data);
```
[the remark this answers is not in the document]
WillSewell: I think it would be nicer to bind to this event from a level above, and log. The reason I didn't was because I didn't want to change the behaviour so that `No callbacks on <channel> for pusher:subscription_error` would no longer be logged. See:

https://github.com/pusher/pusher-js/blob/cac7255980a74977bbe40365fc161579deeb945a/src/core/channels/channel.ts#L27

I guess I could duplicate this log line in the handler for `pusher:subscription_error`, but I'm not sure that's much better.

src/core/channels/encrypted_channel.ts
WillSewell: Isn't the `callback` passed in from the user of the library? The authorizer of the superclass is called further up in this method.

src/core/logger.ts
WillSewell: `defaultLoggingFunction` is the log function to use if the user of the library has not specified anything. Can you think of a better name? If not perhaps a parameter doc would help.

WillSewell:
> Also any reason not to run the formatter?

I thought I had...

Also, do you have any suggestions for how I could implement "private methods" for the Logger object?

src/core/channels/channel.ts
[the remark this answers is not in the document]
leesio: I think that was my thinking. It is a good point that this might make it *less* obvious that you can `bind` to error events.

src/core/channels/encrypted_channel.ts
leesio: I don't think so. The `callback` is passed in by the `subscribe` method of the `Channel` type here:

https://github.com/pusher/pusher-js/blob/error-logging-improvements/src/core/channels/channel.ts#L97

This calls `authorize` on the `EncryptedChannel` which called `authorize` on the `PrivateChannel` etc

src/core/channels/channel.ts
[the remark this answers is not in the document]
WillSewell: I don't feel too strongly about it. I left it as it is. I added a comment to explain why it is logged here.

src/core/channels/encrypted_channel.ts
WillSewell: Ah yes you are correct. I found the flow confusing going from `Channel.subscribe -> EncryptedChannel.authorize -> PrivateChannel.authorize`. Good catch!

Issue #392, opened 2019-10-02
`SameSite` attribute Warning


yofret:
**Do you want to request a *feature* or report a *bug*?**
I think is neither of those

**What is the current behavior?**
I'm getting this warning, starting today, the date I'm creating this issue

> A cookie associated with a cross-site resource at http://pusher.com/ was set without the `SameSite` attribute. A future release of Chrome will only deliver cookies with cross-site requests if they are set with `SameSite=None` and `Secure`. You can review cookies in developer tools under Application>Storage>Cookies and see more details at https://www.chromestatus.com/feature/5088147346030592 and https://www.chromestatus.com/feature/5633521622188032.

**If the current behavior is a bug, please provide the steps to reproduce and
if possible a minimal demo of the problem via https://jsfiddle.net or similar.**

**What is the expected behavior?**
There should not be a warning indicating that in the future the cookie won't be delivered if the attributes are not set correctly

**Which versions of Pusher, and which browsers / OS are affected by this issue?
Did this work in previous versions of Pusher? If so, which?**
```
"pusher-js": "^5.0.0"

Chrome Version 77.0.3865.90 (Official Build) (64-bit)
```

Thanks for reading this weird issue, and apologize in advance if it doesn't meets the format, I would like to know if there is something I could do from the configuration perspective, to remove this warning, I'm using this configuration https://lighthouse-php.com/4.1/subscriptions/client-implementations.html#apollo

daveayerst:
@yofret I am seeing the same issue. We are using pusher-js in an "old-fashioned" html and javascript site.

yofret:
@leesio Hi James!, that is correct, the issue appears in my application which has pusher-js as a dependency.

Looks like chrome is warning us about a new near-future cookie policy for cross-site request, and it is enforcing the usage of the `SameSite` flag.

BrandonMathis:
Yes, this is surely being caused by a recent release of chrome. I am using pusher-js with the pusher rubygem on a rails application. Perhaps a setting needs to be made on the backend side of things? I  am going to continue investigating and will post here if a discover a fix to handle this warning.

leesio:
I had a look at this issue this morning. I was able to see the warning in a normal Chrome window but not in an Incognito window. This, along with the fact that pusher-js doesn't really use cookies directly, leads me to suspect that this is an issue with one of the cookies set by the pusher dashboard, rather than a specific issue with pusher-js.

[...]

PR #404, opened 2019-12-04
Use XHR for stats requests because JSONP sends cookies, resulting in ugly warnings

jameshfisher:
## What does this PR do?

Uses the XHR transport for stats requests from the "web" runtime, replacing JSONP. The JSONP transport unnecessarily sends cookies, which results in ugly warnings. The XHR transport does not.

This should fix https://github.com/pusher/pusher-js/issues/392

Open question: were we using JSONP for other reasons?

## Checklist

- [ ] All tests are passing.

### DIFF
--- a/spec/javascripts/unit/core/pusher_spec.js
+++ b/spec/javascripts/unit/core/pusher_spec.js
@@ -12,18 +12,7 @@ var Runtime = require('runtime').default;
 
 describe("Pusher", function() {
   var _isReady, _instances, _logToConsole;
-
-  switch (TestEnv) {
-    case "worker":
-    case "node":
-      var timelineTransport = "xhr";
-      break;
-    case "web":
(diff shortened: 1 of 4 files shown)


jameshfisher:
I just stumbled on this, which is probably an alternative that lets us keep JSONP: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/crossorigin

jameshfisher:
We've decided to solve this in a different way - by turning reporting off by default.

TeamballDev:
What happens is that the Pusher domains request that the route be verified with an SSL certificate for more security, if it does not block it, after you upload your project to a server with an SSL certificate, everything works normally, meanwhile you can use,https://chrome.google.com/webstore/detail/allow-cors-access-control/lhobafahddgcelffkeicbaginigeejlf?hl=es regards

PR #407, opened 2020-01-08
Make stats opt in

leesio:
## What does this PR do?

[Description here]

## Checklist

- [x] All tests are passing.

### DIFF
--- a/spec/javascripts/unit/core/pusher_spec.js
+++ b/spec/javascripts/unit/core/pusher_spec.js
@@ -504,7 +504,7 @@ describe("Pusher", function() {
       );
     });
 
-    it("should not be sent by defaul", function() {
+    it("should not be sent by default", function() {
       var pusher = new Pusher("foo");
       pusher.connect();
       pusher.connection.options.timeline.info({});
--- a/spec/javascripts/unit/core/pusher_spec.js
+++ b/spec/javascripts/unit/core/pusher_spec.js
@@ -262,7 +262,7 @@ describe("Pusher", function() {
 
   describe("#connect", function() {
     it("should call connect on connection manager", function() {
-      var pusher = new Pusher("foo", { disableStats: true });
+      var pusher = new Pusher("foo");
       pusher.connect();
       expect(pusher.connection.connect).toHaveBeenCalledWith();
     });
@@ -270,14 +270,14 @@ describe("Pusher", function() {
 
   describe("after connecting", function() {
     beforeEach(function() {
-      pusher = new Pusher("foo", { disableStats: true });
+      pusher = new Pusher("foo");
       pusher.connect();
       pusher.connection.state = "connected";
       pusher.connection.emit("connected");
     });
 
     it("should subscribe to all channels", function() {
-      var pusher = new Pusher("foo", { disableStats: true });
+      var pusher = new Pusher("foo");
 
       var subscribedChannels = {
         "channel1": pusher.subscribe("channel1"),
@@ -357,7 +357,7 @@ describe("Pusher", function() {
     var pusher;
 
     beforeEach(function() {
-      pusher = new Pusher("foo", { disableStats: true });
+      pusher = new Pusher("foo");
     });
 
     it("should pass events to their channels", function() {
@@ -452,7 +452,7 @@ describe("Pusher", function() {
 
   describe("after disconnecting", function() {
     it("should disconnect channels", function() {
-      var pusher = new Pusher("foo", { disableStats: true });
+      var pusher = new Pusher("foo");
       var channel1 = pusher.subscribe("channel1");
       var channel2 = pusher.subscribe("channel2");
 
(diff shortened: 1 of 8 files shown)


spec/javascripts/unit/core/pusher_spec.js
```diff
@@ -496,15 +496,16 @@ describe("Pusher", function() {
...
+        enableStats: true,
       });
       expect(Factory.createTimelineSender).toHaveBeenCalledWith(
         pusher.timeline, { host: "example.com", path: "/timeline/v2/" + timelineTransport }
       );
     });
 
-    it("should not be sent if disableStats option is passed", function() {
-      var pusher = new Pusher("foo", { disableStats: true });
+    it("should not be sent by defaul", function() {
```
jameshfisher: unimportant typo `defaul`

jameshfisher:
LGTM, just wanted to check one thing with @pl below vvv

Thinking through backwards-compatibility for this API change:

`new Pusher("key")`: this is backwards-compatible, since the user doesn't care what stats behavior it has.

`new Pusher("key", { disableStats: true })`: this is backwards-compatible; does not change reporting behavior.

`new Pusher("key", { disableStats: false })`: this is arguably a breaking change: the user explicitly requested that it report stats, but now it won't unless they switch to `enableStats: true`. But I believe that very few users will have `disableStats: false` explicitly set, and won't have anything relying on the SDK reporting stats. @pl - would you agree with this?

Also, upgrading adds a deprecation warning. That's fine, but means we should check code snippets to update them.

spec/javascripts/unit/core/pusher_spec.js
leesio:
```suggestion
    it("should not be sent by default", function() {
```

leesio:
@jameshfisher that's a fair point. I hadn't considered that edge case.

pl:
Another option is to keep the `disableStats` flag and use the `in` operator to check whether the key was specified in constructor options (even if it was false). This way you can flip the default while keeping the stats reporting in the third case.

src/core/pusher.ts
```diff
@@ -88,7 +93,7 @@ export default class Pusher {
       level: TimelineLevel.INFO,
       version: Defaults.VERSION
     });
-    if (!this.config.disableStats) {
+    if (this.config.enableStats) {
```
jameshfisher: If we want to preserve backwards compatibility, we could change this to

    if (this.config.enableStats || (this.config.enableStats === undefined && this.config.disableStats === false)) {

src/core/pusher.ts
```diff
@@ -68,6 +68,11 @@ export default class Pusher {
         `You should always specify a cluster when connecting. ${suffix}`
       );
     }
+    if (options.disableStats) {
```
jameshfisher: You might also want a check like

```js
if (options.enableStats !== undefined && options.disableStats !== undefined) {
   Logger.error('Ambiguous: remove deprecated disableStats from your config')
}
```
```

Comment (leesio) [reply src/core/pusher.ts]: Not sure how strongly you feel about this specific error message.

I replicated the existing behavious in 6e53ba4 but this specific error message is missing. I don't feel it's necessary but if you think it is, I can add it.

File: src/core/pusher.ts (hunk 2 above)
Comment (leesio) [reply src/core/pusher.ts]: This should lead to the same behaviour. I think it's nicer

https://github.com/pusher/pusher-js/pull/407/commits/6e53ba4c1b94b8579e610256b57eba6a7138fc2c

File: src/core/pusher.ts [hunk 4]
```diff
@@ -68,6 +68,14 @@ export default class Pusher {
         `You should always specify a cluster when connecting. ${suffix}`
       );
     }
+    if ('disableStats' in options) {
+      Logger.warn(
+        'The disableStats option is deprecated, stats are now disabled by default'
```
jameshfisher: "deprecated in favor of enableStats"?

spec/javascripts/unit/core/pusher_spec.js
```diff
@@ -496,15 +496,16 @@ describe("Pusher", function() {
...
+        enableStats: true,
       });
       expect(Factory.createTimelineSender).toHaveBeenCalledWith(
         pusher.timeline, { host: "example.com", path: "/timeline/v2/" + timelineTransport }
       );
     });
 
-    it("should not be sent if disableStats option is passed", function() {
-      var pusher = new Pusher("foo", { disableStats: true });
+    it("should not be sent by default", function() {
```
jameshfisher: Since we consider the `disableStats: false` case important, perhaps we should add a test case for it, by copying the test case for `enableStats: true`?

src/core/pusher.ts
leesio:
```suggestion
        'The disableStats option is deprecated in favor of enableStats'
```

PR #420, opened 2020-02-11
Bundle ts types

leesio:
## What does this PR do?

Include typescript types

## Checklist

- [ ] All tests are passing.


jordanryanmoore:
I could be wrong, but it looks like this broke things for node. See #439.

PR #428, opened 2020-02-13
fix #427: custom authorizer callback handles string value

### LINKED ISSUE #427: custom authorizer: callback accepts "string" data but fails with it
This is a bug in the Typescript type definition or in its implementation.

Custom authorizer function receive a callback that require two parameters. A boolean value indicating an error and the authentication data. For TypeScript the type of "data" is set to `AuthInfo` (see file [`core/auth/options.ts`](https://github.com/pusher/pusher-js/blob/3e49245daea296f97b8ab439a7a9efa3a4367fa8/src/core/auth/options.ts#L15)).

`AuthInfo` can either be a `string` an object of type `AuthData`. However, the callback can only work with `AuthData` and will fail with type `string` (see file [src/core/channels/channel.ts](https://github.com/pusher/pusher-js/blob/3e49245daea296f97b8ab439a7a9efa3a4367fa8/src/core/channels/channel.ts#L108)).

* use custom authorization function
* provide a `string` value to the callback

The call back should check for string type and then convert the string value to the `AuthData` instance itself.

at least 5.1 is affected, I did not test with earlier versions.

nros:
## What does this PR do?

fixes #427 

The callback passed to custom authorizer functions has not
handled string values yet. The provided authorization
data was just castet to an AuthData object.

Now, the string value should be handled correctly.

## Checklist

- [OK] All tests are passing.
- [ n.a.] New or changed API methods have been documented.
- [OK ] `npm run format` has been run

### DIFF
--- a/src/core/channels/channel.ts
+++ b/src/core/channels/channel.ts
@@ -105,7 +105,7 @@ export default class Channel extends EventsDispatcher {
         Logger.error(data);
         this.emit('pusher:subscription_error', data);
       } else {
-        data = data as AuthData;
+        data = typeof data === 'string' ? { auth: data } : (data as AuthData);
         this.pusher.send_event('pusher:subscribe', {
           auth: data.auth,
           channel_data: data.channel_data,


devdreamsolution (approved):

rh78:
Has this been released already?

PR #429, opened 2020-02-25
Expose type declarations to package consumers

stof:
## What does this PR do?

Type declaration files are now shipped in the package. But the package.json was not declaring them, and so they were not taken into account.


stof:
Actually, the entry point of types is index.d.ts at the root of the package, which is the default location. My testing was wrong.

PR #444, opened 2020-04-21
refactor config

leesio:
## What does this PR do?

Separate Options from Config.

Previously the configuration of the library was very confusing. Configuration
objects were frequently composed by combining multiple different objects with
`Collections.extend`. Types annotations were absent and it made tracing the
config tricky.

There were also default configuration settings dotted all around the codebase
making it hard to see where parameters were set

This PR tries to improve that by separating the Options (an object of optional
parameters passed into the constructor by the user of the library) from the
Config (the full configuration object used by the library). The Config has
default parameters but can be overridden by Options

- create distinct types for the Config (used by the library) and the
Options (argument passed into the constructor allowing the library user
    to override default Config parameters)

- move the config creation all into the config module
- add a default cluster and set it on the config - this means that if no cluster is supplied, clients connect to `ws-mt1.pusher.com` not `ws-pusherapp.com` as previously
- fix tests that expect pusherapp.com host

## Checklist

- [x] All tests are passing.
~- [ ] New or changed API methods have been documented.~ there shouldn't be any

### DIFF
--- a/src/core/transports/transport.ts
+++ b/src/core/transports/transport.ts
@@ -1,6 +1,7 @@
 import Factory from '../utils/factory';
 import TransportHooks from './transport_hooks';
 import TransportConnection from './transport_connection';
+import TransportConnectionOptions from './transport_connection_options';
 
 /** Provides interface for transport connection instantiation.
  *
--- a/spec/javascripts/integration/core/cluster_config_spec.js
+++ b/spec/javascripts/integration/core/cluster_config_spec.js
@@ -101,20 +101,20 @@ module.exports = function(testConfigs) {
     }
 
     var _VERSION;
-    var _channel_auth_transport;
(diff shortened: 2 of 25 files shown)


src/core/transports/transport.ts
```diff
@@ -46,7 +47,8 @@ export default class Transport {
     name: string,
     priority: number,
     key: string,
-    options: any
+    options: any // TODO
```
elverkilde, line 50: can't we track this somewhere else?

elverkilde: (the TODO)

leesio: I'll just remove it, I'm not going to do this as part of this PR now. I'm conscious of trying to make these changes incrementally

PR #445, opened 2020-04-21
allow nacl implementation to be passed as option

leesio:
The bundle size is increases significantly by the inclusion of a nacl
implementation for end to end encryption.

Since not all users of the library will require end to end encryption,
we'd prefer to keep this large dependency out of the main library.

This commit makes it possible for consumers of the library to inject a
nacl implementation into pusher-js like so:

```
const pusher = new Pusher(key, {
  nacl: nacl,
});
```

It's still WIP but there's been a reasonable amount of interest in this
functionality so I'm pushing the PR as a draft

## What does this PR do?

[Description here]

## Checklist

- [ ] All tests are passing.


leesio:
Superceded by #447

PR #458, opened 2020-05-02
Deleted the note for not working RN builds

dmitrybirin:
## What does this PR do?

Updating the docs regarding React Native implementation of encrypted channels.

Debatable, cause I saw changes in the previous commits, but the whole note still presents.
But do the note really still needed? It's plainly saying that encryption channels will not work. But they are actually working even without crypto library onboard.


leesio:
You’re right! This was an oversight on my part. 
Thanks for the PR 😀

PR #459, opened 2020-05-02
deleted dummy tweetnacl libs for RN

dmitrybirin:
## What does this PR do?

Deleted dummy files for old encryption channels implementation for React Native
As I see, it's not needed anymore?

Tested by building new react-native build and try it with encryption channels on our app


leesio:
I think this was also an oversight on my part. 😅

PR #465, opened 2020-05-22
Node wss sni

tauu:
## What does this PR do?

The websocket interface of the nodejs runtime is changed, such that it automatically sets the servername property when creating the connection. This property in turn enables SNI during the connection setup and fixes issue where multiple domains are hosted by on the same IP.

## Checklist

### DIFF
--- a/dist/node/pusher.js
+++ b/dist/node/pusher.js
@@ -10199,7 +10199,7 @@ var xhr_timeline_xhr = {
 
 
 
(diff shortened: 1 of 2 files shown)


leesio:
Thanks for this!

Could you point me in the direction of the change in the websocket api for node? I haven't seen it.

tauu:
Maybe my comment was a bit misleading. The websocket api of nodejs has not changed (at least to my knowledge). This PR just changes the interface for websockets in the nodejs runtime of this package to enable SNI. Relevant change is this commit:

https://github.com/pusher/pusher-js/pull/465/commits/692f0069b9becc6f039a0485f8028eeee48b40a5

devdreamsolution (approved):

Issue #437, opened 2020-03-09
Error in event handlers when handling JSON is caught and discarded 


gmrchk:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
Any error in event handlers when handling JSON is caught and discarded by the library. If the error repeats with the string version of the input, only that is displayed. 
This try-catch is the cause.
https://github.com/pusher/pusher-js/blob/3e49245daea296f97b8ab439a7a9efa3a4367fa8/src/core/channels/encrypted_channel.ts#L123
The library will catch any error in the event handler, and emit the event once again with a string. In case the handler contains any check for the format of input and doesn't run with the string input, it won't error at all. 

In general, being benevolent about the format (try JSON, fallback to string) might not be an optimal solution, but that's just my opinion. 

**If the current behavior is a bug, please provide the steps to reproduce and
if possible a minimal demo of the problem via https://jsfiddle.net or similar.**
Throw an error in any event handler. It will only throw an error in the second run here.
https://github.com/pusher/pusher-js/blob/3e49245daea296f97b8ab439a7a9efa3a4367fa8/src/core/channels/encrypted_channel.ts#L127

```javascript
const handler = () => throw Error("Error that won't be displayed");
channel.bind(`api-${eventName}`, handler);
```

**What is the expected behavior?**
Try-catch block should only try to parse the JSON and another try-catch can be present for the actual call of the handler. 

**Which versions of Pusher, and which browsers / OS are affected by this issue?
Did this work in previous versions of Pusher? If so, which?**
Latest, probably continuous problem.

leesio:
@gmrchk thanks for raising this. This is definitely not good behaviour. If I've understood properly, I think #475 should resolve it. What do you think?

> In general, being benevolent about the format (try JSON, fallback to string) might not be an optimal solution, but that's just my opinion.
I happen to share this opinion. Unfortunately it's a change that has a high chance of breaking lots of stuff so it would have to be in a major release.

gmrchk:
Nice work @leesio. It should save many people tedious debugging by showing the actual error right away. It was very confusing to search for why is there even a string received instead of JSON.

> I happen to share this opinion. Unfortunately it's a change that has a high chance of breaking lots of stuff so it would have to be in a major release.

Yep, definitely something to learn from in the future versions. Your fix addresses the main issue now. 👍

leesio:
[...]

PR #475, opened 2020-07-06
Don't catch and discard errors thrown by handlers

leesio:
## What does this PR do?
As per https://github.com/pusher/pusher-js/issues/437.
If the handler throws an error, we catch it and discard it. We shouldn't be doing anything with the caller's handlers.

## Checklist

- [ ] All tests are passing.

### DIFF
--- a/src/core/channels/encrypted_channel.ts
+++ b/src/core/channels/encrypted_channel.ts
@@ -119,23 +119,22 @@ export default class EncryptedChannel extends PrivateChannel {
           );
           return;
         }
-        this.emitJSON(event, encodeUTF8(bytes));
+        this.emit(event, this.getDataToEmit(bytes));
         return;
       });
       return;
     }
-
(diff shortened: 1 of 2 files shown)


src/core/channels/encrypted_channel.ts
```diff
@@ -129,11 +129,13 @@ export default class EncryptedChannel extends PrivateChannel {
   }
 
   emitJSON(eventName: string, data?: any): Dispatcher {
```
gmrchk, line 131: Would something like `emitFormattedData` make more sense? The product of this function doesn't really have to be JSON.

leesio: That's a fair point.
I moved it around a little in e20f31b. I think that's clearer

gmrchk: Beautiful! 🥇

PR #477, opened 2020-07-08
Update auth errors

### LINKED ISSUE #427: custom authorizer: callback accepts "string" data but fails with it

leesio:
## What does this PR do?
Updates the types/callback signature of authorizers to provide a more idiomatic experience.

Fixes #427 
## Checklist

- [x] All tests are passing.

### DIFF
--- a/src/runtimes/worker/auth/fetch_auth.ts
+++ b/src/runtimes/worker/auth/fetch_auth.ts
@@ -2,6 +2,7 @@ import AbstractRuntime from 'runtimes/interface';
 import Logger from 'core/logger';
 import { AuthTransport } from 'core/auth/auth_transports';
 import { AuthorizerCallback, AuthData } from 'core/auth/options';
+import { HTTPAuthError } from 'core/errors';
 
 var fetchAuth: AuthTransport = function(
   context: AbstractRuntime,
@@ -28,10 +29,10 @@ var fetchAuth: AuthTransport = function(
       let { status } = response;
       if (status === 200) {
         return response.text();
-      } else {
-        Logger.error("Couldn't get auth info from your auth endpoint", status);
-        throw status;
       }
+      let message = `Couldn't get auth info from your auth endpoint, status: ${status}`;
+      Logger.error(message);
+      throw new HTTPAuthError(200, message);
     })
     .then(data => {
       let parsedData: AuthData;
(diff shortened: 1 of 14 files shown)


src/runtimes/worker/auth/fetch_auth.ts
```diff
@@ -28,10 +29,10 @@ var fetchAuth: AuthTransport = function(
       let { status } = response;
       if (status === 200) {
         return response.text();
-      } else {
-        Logger.error("Couldn't get auth info from your auth endpoint", status);
-        throw status;
       }
+      let message = `Couldn't get auth info from your auth endpoint, status: ${status}`;
```
jonheslop: Nit pick this might be a little clearer for some people
```suggestion
      let message = `Could not get auth info from your auth endpoint, status: ${status}`;
```

src/runtimes/worker/auth/fetch_auth.ts
```diff
@@ -27,20 +27,25 @@ var fetchAuth: AuthTransport = function(
...
       if (status === 200) {
+        // manually parse the json so we can provide a more helpful error in
+        // failure case
         return response.text();
       }
-      let message = `Couldn't get auth info from your auth endpoint, status: ${status}`;
-      throw new HTTPAuthError(200, message);
+      throw new HTTPAuthError(
+        200,
+        `Couldn't get auth info from your auth endpoint, status: ${status}`
```
jonheslop, line 36: Same again, negative contractions are harder to understand for some people.
```suggestion
        `Could not get auth info from your auth endpoint, status: ${status}`
```

PR #482, opened 2020-07-21
introduce keepAlive

wille:
## What does this PR do?

Adds the `keepAlive` config option which defaults to the current behaviour, `true`. If set to `false`, it will only have an active connection to pusher if there is any active channel subscription. There will be no connection to pusher if there no channels subscriptions.

## Checklist

- [x] All tests are passing.


devdreamsolution (approved):

wille:
Need some help figuring out why the test wont pass.

wille:
Bump

benw-pusher:
You can achieve something very similar by using private or presence channels and then turning on [authorised connections](https://pusher.com/docs/channels/using_channels/authorized-connections/).

PR #500, opened 2020-10-27
Fix stuck subscription pending state

### LINKED ISSUE #255: Retrying subscription on failed auth for private channel
I'm not sure if this is a bug or just incomplete documentation.

When I subscribe to a private channel, and the auth server fails for some reason, the channel gets stuck in a weird state that I can't retry. Trying pusher.subscribe() on the same channel names returns the existing channel object but doesn't perform an auth request to subscribe to the channel.

if possible a minimal demo of the problem via https://jsfiddle.net or similar.

```
var pusher = new Pusher('APP_KEY');
var channel = pusher.subscribe('private-channel');
channel.bind('pusher:subscription_error', function(status) {
  if(status == 408 || status == 503){
    // retry?
  }
});
```

this is the sample code from the documentation: https://pusher.com/docs/client_api_guide/client_events#subscription_error

but there's no information on the proper method to retry. calling pusher.subscribe() again doesn't do anything.

pusher.subscribe() on a channel that failed to auth should subscribe the channel.

As a work around, I dug through the source and found the channel has a method subscribe() that retries the auth process. However, in the code this method was marked as internal use only, so I can't count on it being there in the future. It does work to retry though. It just looks like the pusher.subscribe() method doesn't handle the case where a channel is in a failed auth state and retries.

TomKemp:
## What does this PR do?

Resolves #255 

Currently if an auth request fails, a channel remains in a "subscriptionPending" state meaning it isn't possible to attempt to subscribe to the channel again. This PR clears the "subscriptionPending" state if the auth request fails, meaning it is possible to call `unsubscribe` and then `subscribe` again, to reattempt the subscription.

In order to do this, I made a slight change to the `unsubscribe` method too, so that the unsubscribe message is only sent over the websocket if the channel is actually subscribed. It appears that previously, while the websocket is connected, a channel could only be in the "subscriptionPending" or "subscribed" state. But now it's possible for channels to be in a state where they aren't subscribed and also aren't pending (if the auth request fails). In that case, we shouldn't send the "unsubscribe" message over the websocket when calling unsubscribe, because we aren't subscribed.

## Checklist

- [x] All tests are passing.

### DIFF
--- a/spec/javascripts/unit/core/channels/channel_spec.js
+++ b/spec/javascripts/unit/core/channels/channel_spec.js
@@ -276,6 +276,19 @@ describe("Channel", function() {
       expect(channel.subscriptionPending).toEqual(false);
     });
 
+    it("should keep #subscriptionPending set as true after a successful authorization", function() {
+      expect(channel.subscriptionPending).toEqual(false);
+
+      channel.subscribe();
+      var authorizeCallback = channel.authorize.calls[0].args[1];
+      authorizeCallback(false, {
+        auth: "one",
+        channel_data: "two"
+      });
+
+      expect(channel.subscriptionPending).toEqual(true);
+    });
+
     it("should do nothing if already subscribed", function() {
       channel.subscribed = true;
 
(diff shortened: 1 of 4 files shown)


leesio (approved):
LGTM. Nice work.

PR #509, opened 2020-11-16
withCredentials auth option

### LINKED ISSUE #99: Add `withCredentials` flag to auth params
See #62 for discussion and a workaround.

### LINKED ISSUE #471: Request: ability to set xmlHttpRequest.withCredentials
*Feature*

`xmlHttpRequest.withCredentials` takes on the default value (`false`) and I can't use Pusher auth calls to set cookies.

Example code:
```typescript
const pusher = new Pusher(REACT_APP_PUSHER_KEY, {
  cluster: REACT_APP_PUSHER_CLUSTER,
  authEndpoint: `${REACT_APP_BACKEND_HOST}/pusher/auth/web`,
  auth: {
    headers: {},
    params: {
        // [irrelevant]
    },
  },
});

pusher.connect();
```

`REACT_APP_BACKEND_HOST` is potentially on a different origin. When the backend sends
```
Set-Cookie: connect.sid=s%3AKjPBOeBYzc705H7vq4qnkSXPHV2zF1qi.x7N8w9e2sCc0Z2hQHeWaBy9b73M04gI0qi0WkY%2F%2BbYc; Path=/; Expires=Tue, 22 Sep 2020 03:11:57 GMT; HttpOnly
```

The cookie is not persisted in the browser.

Expose the internal `xmlHttpRequest` config so I can set `withCredentials = true`. I'm using this in the Web runtime.

This hasn't been an option in the past to my knowledge. I'm using Pusher JS `6.0.3`.

I've seen the workarounds suggested in https://github.com/pusher/pusher-js/issues/62, but I'd rather not have to keep up with private API changes and I think this is a widely applicable use case! 🙂

Thank you so much!

callum-oakley:
## What does this PR do?

fixes #471 

## Checklist

- [ ] All tests are passing.


leesio:
It probably makes sense for this to apply to the fetch based authorizer used in the web worker aswell.

(I'm not expecting you to do this, @callum-oakley 😬 )

PR #518, opened 2020-12-16
give opts.forceTLS priority over current protocol

simonbuehler:
## What does this PR do?

hi,

this change gives the forceTLS Option priority over the current protocol, so when hosting the app on https and the (devel) websocket server is running on http (laravel websockets, laradock) there is no implicit changing of the settings, avoiding things like

`Pusher :  : ["Connecting",{"transport":"ws","url":"wss://grabber.test:6001/app/mykey?protocol=7&client=js&version=7.0.2&flash=false"}]`
## Checklist

- [ ] All tests are passing.

### DIFF
--- a/src/core/config.ts
+++ b/src/core/config.ts
@@ -109,10 +109,10 @@ function getWebsocketHostFromCluster(cluster: string): string {
 }
 
 function shouldUseTLS(opts: Options): boolean {
-  if (Runtime.getProtocol() === 'https:') {
-    return true;
-  } else if (opts.forceTLS === false) {
+  if (opts.forceTLS === false) {
     return false;
+  } else if (Runtime.getProtocol() === 'https:') {
+    return true;
   }
   return true;
 }


leesio:
Thanks for creating a PR for this. I'm not sure the `forceTLS` option is the best place to control this behaviour though.

At the moment it works like

`forceTLS` | protocol | use only TLS?
--- | --- | ---
`true` | `http` | `true`
`true` | `https` | `true`
`false` | `https` | `true`
`false` | `http` | `false`

I think this PR changes that to 

`forceTLS` | protocol | use only TLS?
--- | --- | ---
`true` | `http` | `true`
`true` | `https` | `true`
`false` | `https` | `false`
`false` | `http` | `false`

I don't think it's obvious that setting `forceTLS` to `false` would cause the library to open insecure websocket connections on secure pages! I wonder whether this might be better as another option with a more descriptive name and scary name like `allowInsecureConnectionsOnSecurePages` (just an example)

simonbuehler:
i see your point of keeping the security up but the implicit behaviour of the parameter is misleading at least ( and caused some hours of headache and stepping through code ) and maybe an extra option may be better. It just should be somehow possible to use a non-ssl websocket Server in a ssl enabled Dev Environment

PR #519, opened 2020-12-17
Upgrade faye-websocket to 0.11.3

### LINKED ISSUE #494: Custom host WSS connections fail to connect on NodeJS
Bug

Connecting to a laravel-websockets server in a node.js application fails and state changes to `unavailable`. Server doesn't show any indication of a client connecting. Same config, wss url, etc. works fine in the browser runtime. After some digging, I first thought the fault lies somewhere with `faye-websocket` (only referenced for the Node runtime) and TLS connection errors.

However, weirdly, I am able to fix the problem by replacing [this line](https://github.com/pusher/pusher-js/blob/9152c97631d08d428d36ded807b8c81ef7c008d3/dist/node/pusher.js#L10095) with the following:
```js
var websocket = require('faye-websocket');
```
i.e. bypassing webpack's own require function.

Any ideas as to why the webpack compiled version of faye-websocket is having trouble connecting? The webpack/config.node.js looks fine.

Using version 7.0.0, also tried with 6.0.3. On macOS and Ubuntu as well. Tried with nodejs 14 and 12 as well.

```js
const ps = new Pusher(config.pusherKey, {
  wsHost: config.wsHost,
  enabledTransports: ['ws', 'wss'],
  disableStats: true,
  forceTLS: true,
});
```
This is the config I'm using however I don't think it would affect anything, as it's not passed to the WebSocket client constructor. 

Many thanks!

Akhawais:
## What does this PR do?

Fixes #494 - Upgrades the version of `faye-websocket` to 0.11.3.

## Checklist

N/A - only change is dependency version.

- [ ] All tests are passing.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2169,9 +2169,9 @@
       "dev": true
     },
     "faye-websocket": {
-      "version": "0.9.4",
-      "resolved": "https://registry.npmjs.org/faye-websocket/-/faye-websocket-0.9.4.tgz",
-      "integrity": "sha1-iFk0x57/sECVSeDAo4Ae0XpAza0=",
+      "version": "0.11.3",
+      "resolved": "https://registry.npmjs.org/faye-websocket/-/faye-websocket-0.11.3.tgz",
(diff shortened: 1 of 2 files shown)


leesio:
Thanks for preparing this.

leesio:
@Akhawais I just pushed a release (7.0.3) which includes this change. Thanks a lot for creating the PR

PR #551, opened 2021-07-23
Resolve npm audit issues

gileswells:
## What does this PR do?

Resolves the `critical` npm package advisory among others.

| When | Low | Moderate | High | Critical | Total |
|---|---|---|---|---|---|
| Before | 221 | 9 | 18 | 1 | 249 |
| After | 18 | 4 | 4 | 0 | 26 |

## Checklist

- [ ] All tests are passing.

## Future considerations

This should be seen as a temporary remedy only and the better long term solution would be to properly update the packages and through their parent dependencies. There were some fixes here that would have required major version upgrades to install without sidestepping through the use of `npm-force-resolutions`.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -296,6 +296,14 @@
         "kind-of": "^3.0.2",
         "longest": "^1.0.1",
         "repeat-string": "^1.5.2"
+      },
+      "dependencies": {
+        "kind-of": {
+          "version": "6.0.3",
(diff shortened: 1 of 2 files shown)


amad:
@gileswells most of the issues were resolved by updating dependencies, as you also pointed out https://github.com/pusher/pusher-js/pull/554.

I'm going to close this PR. Please re-open it if you think these changes are still needed.

PR #556, opened 2021-09-27
Document the authEndpoint as being passed to authorizer generators

stof:
## What does this PR do?

This improves type definitions for the AuthorizerGenerator.

In practice, the implementation passes the full `Config` object, which is why the `authEndpoint` config was available.
However, type definitions only define that `AuthorizerOptions` is passed, so it should expose this `authEndpoint` setting that is used by all implementations

## Checklist

- [x] All tests are passing.


stof:
@amad aren't those auto-generated by the build process ?

PR #560, opened 2021-10-18
Create release workflow

amad:
## What does this PR do?

Add a workflow to automate release process.

The workflow triggers on push event on the `master` branch. If and only when
we don't have a tag for the current version, it creates the tag and GitHub release
and then it publishes to the NPM and uploads the build to S3.

image

## Checklist

- [x] All tests are passing.
- [x] N/A ~New or changed API methods have been documented.~
- [x] N/A ~`npm run format` has been run~

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,112 @@
+on:
+  push:
+    branches: [ master ]
+
+jobs:
+  check-release-tag:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+        with:
+          fetch-depth: 0
+      - name: Prepare tag
+        id: prepare_tag
+        continue-on-error: true
+        run: |
+          export TAG=v$(jq -r '.version' package.json)
+          echo "TAG=$TAG" >> $GITHUB_ENV
+
+          export CHECK_TAG=$(git tag | grep $TAG)
+          if [[ $CHECK_TAG ]]; then
+            echo "Skipping because release tag already exists"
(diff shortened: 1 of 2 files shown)


.github/workflows/release.yml
```diff
@@ -0,0 +1,112 @@
...
+          aws s3 sync dist/web/ s3://${BUCKET_NAME}/${VERSION}/ --acl public-read --cache-control max-age=$MAX_AGE
+          aws s3 sync dist/worker/ s3://${BUCKET_NAME}/${VERSION}/ --acl public-read --cache-control max-age=$MAX_AGE
+
+          aws s3 sync dist/web/ s3://${BUCKET_NAME}/${VERSION_MIN}/ --acl public-read --cache-control max-age=$MAX_AGE
+          aws s3 sync dist/worker/ s3://${BUCKET_NAME}/${VERSION_MIN}/ --acl public-read --cache-control max-age=$MAX_AGE
+
+          aws cloudfront create-invalidation --distribution-id ${CLOUDFRONT_DISTRIBUTION_ID} --paths "/latest/*" "/${VERSION}/*" "/${VERSION_MIN}/*"
+        env:
+          MAX_AGE: 2592000
+          BUCKET_NAME: pusher-js-cloudfront-staging
```
fbenevides, line 108: always sending to staging bucket? is that correct?

fbenevides (approved):

.github/workflows/release.yml
amad: Oops! I forgot to change that. Thanks for flagging this.

PR #600, opened 2022-06-01
Fix exported TS types

### LINKED ISSUE #598: When trying to build an angular project in prod it cannot build due to missing classes
Report a bug:

When trying to use ng build --prod with Pusher-js npm installed I get 3 errors all Authentication related. Tried to import them into 
App.module.ts

That the ng-build --prod will exectute succesfully 
Latest version of pusher, not too sure if this worked on previous versions. 
image_2022-05-24_184453828
image
image

MeenaAlfons:
## What does this PR do?

- Fix exported TS types in index.d.ts by removing deleted types and addinng new types.
- Regenerating `types` directory from scratch. That results in removing deleted files.

Fixes #598

## Checklist

- [x] All tests are passing.


agatav (approved):

sonologico (approved):

index.d.ts
```diff
@@ -1,8 +1,13 @@
 export {
-  Authorizer,
+  DeprecatedAuthOptions,
+  ChannelAuthorizerGenerator,
+} from './types/src/core/auth/deprecated_channel_authorizer';
+export {
   AuthOptions,
-  AuthorizerGenerator,
```
stof, line 4: Removing AuthorizerGenerator breaks https://github.com/dirkbonhomme/pusher-js-auth. Was this really removed in a minor version rather than in a major version ? Or is this just a mistake in the type declarations ?

stof: Looking at the diff, it seems like this was renamed to `ChannelAuthorizerGenerator`. But this is a BC break (as it effectively removes the old type and adds a new one).

MeenaAlfons: That's definitely a mistake. We'll fix it ASAP. Thanks for reporting it. 🙏

MeenaAlfons: Fixed in https://github.com/pusher/pusher-js/pull/615

PR #620, opened 2022-08-05
Bump typescript and @types/node versions

fbenevides:
## What does this PR do?

* Bump typescript version
* Bump @types/node version
* Bump jasmine 
     * It fixes this bug on CI: ` Uncaught exception: TypeError: Cannot read property 'removeFunctionWithId' of null`

## Checklist

- [x] All tests are passing.


amad (approved):

package.json
```diff
@@ -63,6 +63,7 @@
     "xmlhttprequest": "^1.8.0"
   },
   "dependencies": {
-    "tweetnacl": "^1.0.3"
+    "tweetnacl": "^1.0.3",
+    "@types/node": "^14.14.31"
```
stof, line 67: shouldn't that be a dev dependency ? and if it stays a regular dependency, shouldn't it allow a wider range ?

PR #629, opened 2022-08-23
Update release_pr workflow to not pull from private repo

amad:
Update release_pr workflow to not pull from private repo cos that won't
work on external PRs.

The release_pr doesn't work on external PRs because it depends on an
action from a private repo (pusher/actions). External PRs can't use
GitHub Secrets to run Actions which is required in this case to copy an
action from a private repo.

Related to https://github.com/pusher/public_actions/pull/1

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -8,3 +8,7 @@
 - [ ] All tests are passing.
 - [ ] New or changed API methods have been documented.
 - [ ] `npm run format` has been run
+
+## CHANGELOG
+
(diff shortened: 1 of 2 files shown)


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -8,3 +8,7 @@
 - [ ] All tests are passing.
 - [ ] New or changed API methods have been documented.
 - [ ] `npm run format` has been run
+
+## CHANGELOG
+
```
singhashmeet, line 13: Please update it to reflect current change

amad: This is the template!

singhashmeet (approved):

PR #636, opened 2022-10-09
Fixed default value: channelAuthorization.endpoint

omril321:
## What does this PR do?

I've noticed that the documentation for the default value of `channelAuthorization.endpoint` is not correct. It says that the default value is `/pusher/user-auth`, while it looks like it should be `/pusher/auth`. This PR fixes it.

## Checklist

- [ ] All tests are passing.

## CHANGELOG

- [FIXED] Fixed documentation for default value of `channelAuthorization.endpoint`


sonologico (approved):
Thanks!

Issue #685, opened 2023-01-16
HTTP 201 is the correct response code for the authorization route


rharriso:
This bug has been reported twice. Once in  [2012](https://github.com/pusher/pusher-js/issues/17) and again in [2016](https://github.com/pusher/pusher-js/issues/154). The responses both times were essentially "201 is not the correct response code". This is against consensus.

* The MDN docs specifically call out successful `POST` requests as resulting in `201` responses.
* Multiple web frameworks use `201` as the status code for successful `POST` requests.

The justification in 2012 doesn't quite work for me. https://github.com/pusher/pusher-js/issues/17#issuecomment-76370378

> We're not really creating a resource in the auth call, so it can't be referenced later.

Even if that is true (which I doubt since subsequent calls with the same input yield different responses), it's an implementation detail of the pusher backing service. It is not the semantic of API implementing the authentication route. 

When a node application calls `pusher.authorizeChannel` using the server library it appears to that a new authentication object was created. 

----

**Do you want to request a *feature* or report a *bug*?**

Bug

**What is the current behavior?**

Fails unless server returns 200

**What is the expected behavior?**

Accept 201 as a valid response code for the authentication request.

**Which versions of Pusher, and which browsers / OS are affected by this issue?

All of them.

benw-pusher:
Thanks for raising this, I will escalate this internally for review.

rharriso:
Thanks @benw-pusher . For more context this caused an issue when integrating with nest js

benw-pusher:
Reopening as this is still pending review.

PR #686, opened 2023-01-16
Accept HTTP 201 response codes

rharriso:
## What does this PR do?

Accept HTTP response code 201. See issue #685 

## Checklist

- [ ] All tests are passing.

## CHANGELOG

- [FIXED] Support response code `201` from successful `POST` requests

### DIFF
--- a/src/runtimes/worker/auth/fetch_auth.ts
+++ b/src/runtimes/worker/auth/fetch_auth.ts
@@ -39,7 +39,7 @@ var fetchAuth: AuthTransport = function(
   return fetch(request)
     .then(response => {
       let { status } = response;
-      if (status === 200) {
+      if (status === 200 || status === 201) {
         // manually parse the json so we can provide a more helpful error in
         // failure case
         return response.text();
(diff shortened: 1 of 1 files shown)


amad:
@rharriso `pusher.authorizeChannel` shouldn't result a `201`. I couldn't reproduce it. Could you please share a code snippet?

If you are using an open source web application framework, please share its name and version.

rharriso:
Not supporting 201 is a defect. It's not necessary to enumerate all of the frameworks that use 201 for successful authentication.

* `201 Created` is a more specific success response than simply "OK".
* It is very common for frameworks or libraries to define logins as "creating" a session or token or whatever is being used to say that a given user is logged in right now.

Requiring the response code to be 200 necessitated an adapter in the system I was working on 6 months ago. I can't remember if I had to intercept the response on the server or the client, and I no longer have access to that project so I can't confirm.

I'm no longer working with PusherJS, so it doesn't matter to me if this defect is fixed, but it is a defect.

KayakinKoder:
Hasura is an example of a service which returns a 201 (for pretty much all successful api response use cases).

Pusher not allowing the 201 was one of those multi-hour head scratchers that hopefully future users won't have to go through.

PR #733, opened 2023-06-13
Remove retry limit for websocket

amad:
## What does this PR do?

This commit removes the limit on WebSocket failures. Previously, the code set a maximum number of failures before falling back to other transports and using them for all subsequent re-connections. The purpose of this limit was to improve re-connection time for clients experiencing repeated issues with WebSocket transport.

However, it is now believed that this limit is unnecessary. Today, the occurrence of such issues is less common, and the limit may have unintended consequences. For example, in the current approach, a modern web browser with a temporary bad internet connection could exhaust its chances to use WebSocket and be left with slower fallbacks for the remainder of the session.

By removing this limit, we enable indefinite retries for WebSocket. The fallback behavior is unchanged and we still retry fallback transports.

This behavior is extensively explained in the following blog posts.
- https://pusher.com/blog/how-we-built-pusher20-part-1/
- https://pusher.com/blog/how-we-built-pusher-js-2-0-part-2-implementation/
- https://pusher.com/blog/how-we-built-pusher-js-2-0-part-3-metrics/

## CHANGELOG

- [CHANGED] Remove WebSocket retry limit.


fbenevides (approved):

amad:
Before: After two expected failure, we remove WebSocket from transports pool and continue with fallback strategies.
After: We do not remove WebSocket from transports pool regardless of how many failures we detect. This doesn't change fallback strategy though. We still try both WebSocket ans other fallback transports when last successful transport doesn't work.

PR #738, opened 2023-06-30
Update cached strategy to prioritize websocket

amad:
## What does this PR do?

Usually, we immediately retry cached transport, which works well in most cases. However, if a client ends up with fallback transports due to a temporary connection problem, they remain with fallback for the rest of their session.

This change updates the Cached Strategy to prioritize the WebSocket transport. Initially, we always try WebSocket first, even if the cached transport is, for example, long-polling. Each time we override the cached transport and connect to WebSocket, we increment the attempts value. After a few attempts, we stop prioritizing WebSocket and switch to the cached transport immediately.

This change allows clients to have more chances of using WebSocket. But if we determine that fallback transports perform better in their environment, we will continue reconnecting them as quickly as possible to the cached transport.

## CHANGELOG

- [CHANGED] Update cached re-connect strategy to prioritize WebSocket


src/core/strategies/websocket_prioritized_cached_strategy.ts
```diff
@@ -47,7 +48,7 @@ export default class CachedStrategy implements Strategy {
     var strategies = [this.strategy];
     if (info && info.timestamp + this.ttl >= Util.now()) {
       var transport = this.transports[info.transport];
-      if (transport) {
+      if (transport && info.attempts && info.attempts > 3) {
```
amad, line 51: I'm hardcoding a few values here and below. I thought we should get feedback and then maybe later try to refactor it.

MeenaAlfons (approved):

PR #739, opened 2023-07-11
Fix error handling event parameter

fbenevides:
## What does this PR do?

Change doc accordingly to the real behavior

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ One reason this connection might fail is your account being over its' limits. Yo
 ```js
 const pusher = new Pusher('app_key', { cluster: APP_CLUSTER });
 pusher.connection.bind( 'error', function( err ) {
-  if( err.error.data.code === 4004 ) {
+  if( err.data.code === 4004 ) {
     log('Over limit!');
   }
 });


MeenaAlfons (approved):

russellgilbert:
While I'm glad to see this fixed, I wanted to point out that by only changing the README.md file, it now matches getCloseError() in src/core/connection/protocol/protocol.ts, but it no longer matches the [Connection](https://pusher.com/docs/channels/using_channels/connection/#detecting-connection-limits) page of the Pusher Docs.

Also, my guess is that there are already a lot of people that have implemented this error handler as it was originally documented, but never actually tested it. By changing the documentation instead of changing the actual code, it now means that all of those error handlers will remain broken unless they read the release notes and update it. It seems to me that changing the code to match the very old documentation (on two pages) might be better since it would instantly fix everyone's code, except for the few people like me that actually discovered the problem.

Or maybe as a compromise, the function could be changed to store the error code and message in both the old and new locations (`err.data` and `err.error.data`). It wouldn't be a lot of data and it would instantly fix everyone's code regardless of whether they'd found the problem or not.

PR #756, opened 2023-09-01
Add ConnectionState type

ahkhanjani:
A proposal for adding the union of connection states according to [this docs page](https://pusher.com/docs/channels/using_channels/connection/#available-states).

## What does this PR do?

Adds the union type for all the different connection status strings. There could be more places that this is used. This is only a demonstration of my request.

## Checklist

- [ ] All tests are passing.

## CHANGELOG

- [CHANGED] Describe your change here. Look at CHANGELOG.md to see the format.


ahkhanjani:
This will add better typescript support.

PR #773, opened 2023-11-08
Expose a way to switch from one cluster/app to the other

kamalbennani:
## What does this PR do?

This PR introduces a way to tell Pusher client to switch the traffic from one cluster/app to the other.

This is super useful especially in case where Pusher is experiencing some outages allowing you to switch the traffic to a stable cluster without having to deal with resubscribing yourself all the channels and bind all the callbacks to the right events.

## Usage

```ts
const pusher = new Pusher('mt1Key', { cluster: 'mt1' });
const channel = pusher.subscribe('private-channel');
channel.bind('event', console.log);

// Detect pusher outage
pusher.switchCluster({ appKey: "us3Key", cluster: "us3" });

// Here you should expect that all channels and bindings are still working as expected.
```

## Checklist

- [x] All tests are passing.

## CHANGELOG

- [Added] Introduce a new `.switchCluster` method to switch the Pusher client to a different cluster and re-establish all existing subscriptions and channel bindings.

### DIFF
--- a/spec/javascripts/unit/core/pusher_spec.js
+++ b/spec/javascripts/unit/core/pusher_spec.js
@@ -307,6 +307,47 @@ describe("Pusher", function() {
       });
     });
 
+    describe("switch cluster", function() {
+      var pusher;
+      var subscribedChannels
+
+      beforeEach(function() {
+        pusher = new Pusher("foo", {cluster: "mt1"});
+
+        subscribedChannels = {
+          channel1: pusher.subscribe("channel1"),
+          channel2: pusher.subscribe("channel2")
+        };
+
+        pusher.connect();
+        pusher.connection.state = "connected";
+        pusher.connection.emit("connected");
+      });
+
+      it("should resubscribe to all channels", function() {
+        expect(subscribedChannels.channel1.subscribe).toHaveBeenCalledTimes(1);
+        expect(subscribedChannels.channel2.subscribe).toHaveBeenCalledTimes(1);
+
+        pusher.switchCluster({ appKey: 'bar', cluster: 'us3' });
+        pusher.connect();
+        pusher.connection.state = 'connected';
+        pusher.connection.emit('connected');
+
+        expect(subscribedChannels.channel1.subscribe).toHaveBeenCalledTimes(2);
+        expect(subscribedChannels.channel2.subscribe).toHaveBeenCalledTimes(2);
+      });
+
+      it("should send events via the connection manager", function() {
+        pusher.switchCluster({ appKey: 'bar', cluster: 'us3' });
+        pusher.send_event("event", { key: "value" }, "channel");
+        expect(pusher.connection.send_event).toHaveBeenCalledWith(
+          "event",
+          { key: "value" },
+          "channel"
+        );
+      });
+    })
+
     describe("#unsubscribe", function() {
       it("should unsubscribe the channel if subscription is not pending", function() {
         var channel = pusher.subscribe("yyy");
--- a/src/core/pusher.ts
+++ b/src/core/pusher.ts
@@ -1,6 +1,5 @@
 import AbstractRuntime from '../runtimes/interface';
 import Runtime from 'runtime';
-import Util from './util';
 import * as Collections from './utils/collections';
 import Channels from './channels/channels';
 import Channel from './channels/channel';
@@ -10,14 +9,11 @@ import TimelineSender from './timeline/timeline_sender';
 import TimelineLevel from './timeline/level';
 import { defineTransport } from './strategies/strategy_builder';
 import ConnectionManager from './connection/connection_manager';
-import ConnectionManagerOptions from './connection/connection_manager_options';
 import { PeriodicTimer } from './utils/timers';
 import Defaults from './defaults';
-import * as DefaultConfig from './config';
 import Logger from './logger';
 import Factory from './utils/factory';
-import UrlStore from 'core/utils/url_store';
-import { Options, validateOptions } from './options';
+import { Options, ClusterOptions, validateOptions } from './options';
 import { Config, getConfig } from './config';
 import StrategyOptions from './strategies/strategy_options';
 import UserFacade from './user';
@@ -53,6 +49,7 @@ export default class Pusher {
 
   /* INSTANCE PROPERTIES */
   key: string;
+  options: Options;
   config: Config;
   channels: Channels;
   global_emitter: EventsDispatcher;
@@ -139,6 +136,19 @@ export default class Pusher {
     if (Pusher.isReady) {
       this.connect();
     }
+  }
+
+  /**
+   * Allows you to switch Pusher cluster without
+   * losing all the channels/subscription binding
+   * as this is internally managed by the SDK.
+   */
+  switchCluster(options: ClusterOptions) {
+    const { appKey, cluster } = options;
+    this.key = appKey;
+    this.options = { ...this.options, cluster };
+    this.config = getConfig(this.options, this);
+    this.connection.switchCluster(this.key);
   }
 
   channel(name: string): Channel {
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "pusher-js",
-  "version": "8.2.0",
+  "version": "8.3.0",
   "lockfileVersion": 2,
   "requires": true,
   "packages": {
     "": {
       "name": "pusher-js",
-      "version": "8.2.0",
+      "version": "8.3.0",
       "license": "MIT",
       "dependencies": {
         "tweetnacl": "^1.0.3"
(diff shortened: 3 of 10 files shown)


src/core/pusher.ts
```diff
@@ -116,26 +132,46 @@ export default class Pusher {
...
-    Pusher.instances.push(this);
-    this.timeline.info({ instances: Pusher.instances.length });
+  /**
+   * Allows you to switch Pusher cluster without
+   * losing all the channels/subscription binding
+   * as this is internally managed by the SDK.
+   */
+  switchCluster(options: ClusterOptions) {
+    // Automatically disconnect the current connection
+    this.disconnect();
```
kamalbennani, line 162: I was wondering if "should" persist the previous WebSocket connection until the new connection is established, or it matters less? 
Supporting this use-case might add a bit of complexity to the code to support two instances of  `ConnectionManager`.
I'd love to have your feedback.

src/core/pusher.ts
```diff
@@ -116,26 +132,46 @@ export default class Pusher {
...
+  }
 
-    Pusher.instances.push(this);
-    this.timeline.info({ instances: Pusher.instances.length });
+  /**
+   * Allows you to switch Pusher cluster without
+   * losing all the channels/subscription binding
+   * as this is internally managed by the SDK.
+   */
+  switchCluster(options: ClusterOptions) {
```
kamalbennani, line 160: As you can see, we only allow to override the appKey and cluster, but of course this could be extend to the other config options but I guess this could be enabled once we have real usecases?

src/core/pusher.ts
```diff
@@ -116,26 +132,46 @@ export default class Pusher {
...
+    // Automatically disconnect the current connection
+    this.disconnect();
 
-    this.user = new UserFacade(this);
+    const { appKey, cluster } = options;
```
kamalbennani, line 164: Should I manually validate the shape of the options as you're doing on the constructor?

src/core/pusher.ts
fbenevides: Let's just expose appKey and cluster for now and enable other options when we have real use cases.

src/core/pusher.ts
```diff
@@ -116,26 +132,46 @@ export default class Pusher {
...
+
+    this.key = appKey;
+    this.config = { ...this.config, cluster };
```
MeenaAlfons, line 167: `getConfig(opts)` needs to be called to properly setup the `this.config`. Changing the `cluster` is not enough.

This means the original `options` need to be stored, altered with new cluster and key, then passed to `getConfig(opts)` to get new config.

src/core/pusher.ts
```diff
@@ -116,26 +132,46 @@ export default class Pusher {
...
+
+    // Recreate a new timeline instances and connection
+    // for the new tuple (appKey, cluster)
+    this.setupTimeline();
+    this.setupConnection();
```
MeenaAlfons, line 173: A new `UserFacade(this)` need to be created to make use of and bind to the new connection.

MeenaAlfons (changes requested):
Thanks for the PR @kamalbennani. I've added some comments pointing at necessary changes for the `pusher` object to be fully functional after `switchCluster` is called.

MeenaAlfons:
I think this PR needs unit testing to verify that all functionalities are working after `switchCluster` is called.

src/core/pusher.ts
kamalbennani: Good catch!

kamalbennani:
Hi @MeenaAlfons Thanks for the review, I've pushed a commit to resolve all your comments.
Let me know if additional changes are needed.

src/core/pusher.ts
```diff
@@ -116,26 +134,49 @@ export default class Pusher {
...
+    const { appKey, cluster } = options;
+
+    this.key = appKey;
+    this.options = { ...this.options, cluster };
+    this.config = getConfig(this.options, this);
+
+    // Recreate a new timeline instances and connection
+    // for the new tuple (appKey, cluster)
+    this.setupTimeline();
+    this.setupConnection();
```
MeenaAlfons, line 175: At this point, some listeners are still bound to the old connection. All listeners on the old connection needs to be unbound to:
- Alleviate any side effects
- Prevent memory leaks

A side note:
The fact that we will forcefully remove all listeners from the old connection here reveals the danger of implementing `switchCluster` in this way because how would you know which objects/actors need such listeners in order to ask them to bind to reinitialize themselves and bind to the new connection? Examples include `UserFacade` which we know we need to reinitialize just because we looked at it's code and found that it binds to the old connection.
The point of this paragraph is to reveal that fact that the code is stateful in terms of binding listeners to the connection. Creating a new connection in `pusher.connnection` is not enough to guarantee consistency.


Given that note:
- It's crucial to have unit tests to test all functionalities after `switchCluster` is called
- I'd rather implement `switchCluster` in a way that maintains the invariants without the need to look inside each object to make sure it still works.
- Or change the code to make it agnostic to the connection change. For example, introducing an intermediate EventEmitter as a connection delegate all objects bind to and whenever there is a new connection, the new connection will be wired with the delegate while the old connection is unwired from the delegate.

MeenaAlfons (changes requested):

src/core/pusher.ts
kamalbennani: Thanks for detailed comment.
From my understanding, calling `this.connection.disconnect` will result in closing the WebSocket connection as well as freeing up everything bound to the that connection.

Now, there is a valid concern, as `connection` is exposed to the outside world, we could still have some event listeners that are outside of the Pusher SDK, we could either loop through them and bind them again to the same connection or just "unbind_all" to avoid any memory leaks and move that responsibility to the consumer.

I'd still challenge the reason behind this choice of making internal resources available for external usage, but that's a bit outside of this MR scope.

I'm just afraid that we are going too deep in some assumptions and making this PR handle things which are outside of the scope. would you be conformable having Pusher team working on ensuring that those assumptions are verified and maybe create new PR to complete the "missing parts" of this implementation?

I just came up with an idea, what if, we don't create a new `ConnectionManager` instance, but instead update its internal config and trigger a `reconnect` (like if the network went from offline to online) this way we won't need to unbound any events and everything would work as expected?

What do you think?

kamalbennani: I think I came up with something even more simpler and would answer your concerns as it uses literally the same technique as when the Network status changes from offline to online. I will commit shortly.

MeenaAlfons: That's definitely a better approach. The ConnectionManager is the delegate in this case that redirects signals between the active connection and the application.

I think the new implementation is sound. You may even revert the unnecessay refactoring to make the PR even simpler. I think it's still important to run unit tests after switchCluster to make sure all feature will still work.

kamalbennani:
Hi @MeenaAlfons, the PR is ready for another round of review.

kamalbennani:
Hi @MeenaAlfons, would you mind reviewing the PR a second time, highly appreciate it!

spec/javascripts/unit/core/pusher_spec.js
```diff
@@ -307,6 +307,47 @@ describe("Pusher", function() {
...
+      });
+
+      it("should resubscribe to all channels", function() {
+        expect(subscribedChannels.channel1.subscribe).toHaveBeenCalledTimes(1);
+        expect(subscribedChannels.channel2.subscribe).toHaveBeenCalledTimes(1);
+
+        pusher.switchCluster({ appKey: 'bar', cluster: 'us3' });
+        pusher.connect();
+        pusher.connection.state = 'connected';
+        pusher.connection.emit('connected');
```
MeenaAlfons, line 334: I think after calling `switchCluster` the connection should automatically be reconnected (going to `connecting` state then to `connected` state) without calling `pusher.connect()`. So basically, the test should not have `pusher.connect()` after `pusher.switchCluster()`.

MeenaAlfons (commented):
Sorry for the late reply.

Upon running tests I got some errors:
```
1) Pusher after connecting switch cluster should resubscribe to all channels
  - TypeError: this.connection.switchCluster is not a function

2) Pusher after connecting switch cluster should send events via the connection manager
  - TypeError: this.connection.switchCluster is not a function
```
----
In general, I think the current tests try to mock the behavior or reconnecting by manually calling `pusher.connect()` and `pusher.connection.state = 'connected'`, etc. This is not helpful when testing `switchCluster` because the expected behaviour we'd like to test is that the connection will be disconnected and reconnected including all the side effects of reconnection like reinitiating channel authorization and user authentication.

It would be nice if you can make such tests. I could be able to help but it will depend on my availability.

kamalbennani:
Hi @MeenaAlfons, I pushed some missing changes which would make the tests pass! sorry for that.

Regarding your comment, I agree with you in theory, and I'd like to know how those tests are done today when a network switch happens, as it's the same logic that is done here.

If that's not cover, I'd prefer if this could be managed by yourself or a pusher member in phase 2 if that's fine with you?

Thanks.

MeenaAlfons:
Hi @kamalbennani,

I've changed the base branch to `releases/v8.4.0-rc1`. The intention is to deploy a release candidate. This will allow you to start using the new version if you need. This will also make it easier to manually test it.

I still don't know when I'll be able to come around to check what needs to be done for tests to cover these changes. I think a release candidate may unblock you.

Thank you.

MeenaAlfons (approved):

PR #819, opened 2024-05-03
For SSR compatibility, add CJS build and guard global variable access

### LINKED ISSUE #624: Pusher with encryption, window is not defined on Next.js
Bug
```ts
import Pusher from "pusher-js/with-encryption";
```
- Create a next app
- Install pusher-js@latest
- import Pusher from "pusher-js/with-encryption";
- Will crash about window is not defined

Should handle if the lib is being imported in SSR and just early return.

7.3.0 - latest

AlexVipond:
Closes #624

## What does this PR do?

For compatibility with frameworks like NextJS that do server-side rendering, fix two things:
1. Add a CJS build, which users can import as `import Pusher from 'pusher-js/cjs'` or `import Pusher from 'pusher-js/cjs/with-encryption'`
2. Add guard clauses so that global variables like `window` and `document` are not accessed in server environments, where they don't exist.

These two changes allow SSR framework users to import Pusher with a reasonably industry-standard DX, without breaking their apps or resorting to dynamic imports.

## Checklist

- [ ] All tests are passing.
- [X] `npm run format` has been run

## CHANGELOG

- [FIXED] Makes Pusher's browser builds compatible with server-side rendering frameworks


AlexVipond:
Within the last 60 days, people experienced the problem solved by this PR.

I suspect [this](https://github.com/pusher/pusher-js/issues/624#issuecomment-2092339055) is also still a widespread issue among Pusher customers:

> Also worth noting for anyone in this thread: if you wrote `import Pusher from 'pusher-js'`, you're importing the NodeJS build (`dist/node/pusher.js`) instead of the browser build (`dist/web/pusher.js`), because that's where the package.json sends you. If you switch to `import Pusher from 'pusher-js/dist/web/pusher'` to force the browser build, you'll get the same `window is not defined` error from the UMD format.

TheTourer:
This is still a problem for our team—still using [this smart but not ideal workaround](https://github.com/pusher/pusher-js/issues/624#issuecomment-2557404765) wrapped in `useEffect`, which adds unnecessary overhead to our Next.js application. Our team has also filed at least one support request for this. Possible to get some internal Pusher engineer eyes on this PR?

PR #977, opened 2026-09-08
Omit default ports from connection URLs

pataar:
PR description is generated by Claude. But the problem is a real occurring one on our React Native iOS case.

## What does this PR do?

Connection URLs always include the port, e.g. `wss://ws-mt1.pusher.com:443/app/...`. Some WebSocket clients copy an explicit URL port verbatim into the `Host` header. SocketRocket (React Native on iOS) does this unconditionally (`_SRHTTPConnectMessageHost`), producing `Host: ws-mt1.pusher.com:443`. That is valid per RFC 9110 but redundant, and some firewalls and proxies reject it, so the handshake fails on iOS while browsers, which normalise default ports away, are fine.

This drops the port from the URL when it equals the scheme default (443 for TLS, 80 otherwise) in `getGenericURL`, the single place every transport URL (ws/wss/http/https/sockjs, all runtimes) is built. Explicitly configured non-default ports (`wssPort: 1999`, …) are unchanged.

Existing specs asserting the default URLs were updated to the portless form. The custom-port specs already cover the "non-default port is kept" case.

## Checklist

- [x] All tests are passing.
- [ ] New or changed API methods have been documented. (no API change)

## CHANGELOG

- [FIXED] Omit default ports (443/80) from connection URLs so WebSocket clients that copy the URL port into the `Host` header (e.g. SocketRocket on iOS / React Native) no longer send `Host: host:443`, which some firewalls and proxies reject


pataar:
Happy to discuss any changes, or questions

